PR #21, opened 2014-06-29
Create `add_json` shortcut method.

gabrielpjordao:
Same as calling `responses.add` assuming that the body is
a dict and the content type is application/json.


ekohl:
Add now has a ```json=``` parameter that achieves the same, so I think this can be closed.

gabrielpjordao:
Yep.. `json` parameter already provides the same functionality. I'm closing this issue..

PR #45, opened 2014-12-18
Raise if requests are not performed as expected

ticosax:
I expect from my testing tools to detect inconsistencies and unpredictable behaviour.
This is why I would like that `responses` helps me to detect request that has not been performed unlike it should.

I implemented it in a way that is backward compatible, but I'm in favour of enabling this behaviour by default.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -103,6 +103,47 @@ A response can also throw an exception as follows.
     # All calls to 'http://twitter.com/api/1/foobar' will throw exception.
 
 
+Responses as a context manager
+------------------------------
+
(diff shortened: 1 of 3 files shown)


thedrow:
Looks really good.
I think it should be on by default.

dcramer:
I agree this changes looks good. I want to clean up some language, but I'll try to get this merged in later this week.

ticosax:
@dcramer I can simplify the implementation if you agree to introduce this incompatible backward change.
up to you.

PR #46, opened 2015-01-19
Feature: cookie handling

OldhamMade:
- added cookie support

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -32,6 +32,8 @@ else:
 import inspect
 from collections import namedtuple, Sequence, Sized
 from functools import update_wrapper
+from cookies import Cookies
+from requests.utils import cookiejar_from_dict
 from requests.exceptions import ConnectionError
 try:
     from requests.packages.urllib3.response import HTTPResponse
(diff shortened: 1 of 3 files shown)


dcramer:
This looks good. One request, can you remove the version bump in the change? If not eventually I'll get around to changing that and merging this in.

OldhamMade:
Out of interest, why remove the version bump? Functionality is changing enough that some users may want to stay with 0.3.0 so that there are no significant changes to tests (and usage), while others may wish to force the change in deployment which wouldn't necessarily be seen by automated systems (eg. I have 0.3.0, therefore I don't need to do anything).

dcramer:
@OldhamMade its something that generally doesnt make sense with a feature change. i.e. versioning should come outside of changing features.

Generally what I tend to do in my projects is either bump the version immediately after a release, or bump it before. I usually side w/ the latter as then it's easier to decide what next version makes sense (i.e. how big are the changes).

That said, the version doesnt reflect whats in master. 0.3.0 (assuming that was the last release) is tagged, and pushed to pypi, and doesnt change. That means leaving this at 0.3.0 has no affect on anyone.

OldhamMade:
Makes sense!

OldhamMade:
Done.

dcramer:
Thanks!

dcramer:
@OldhamMade quick question about the change. Is this designed to send the cookies with followup requests as well? I'm looking at another pending patch I have that cause a conflict here so I want to understand the behavior better.

OldhamMade:
Yes, it should be persisting cookies between requests, so that one can mimic sessions.

dcramer:
@OldhamMade alright thats good to know. I have a patch that actually changes this to use sessions, so I'll get a test worked up and hopefully be able to simply remove that code.

OldhamMade:
@dcramer Sounds good, but don't forget that some users may wish to mimic cookies without sessions, e.g. preferences stored on the client-side, rather than server-side.

FodT:
@dcramer the last few comments describe the trouble I've run into in issue #80 - do you have anything to share?

PR #51, opened 2015-03-06
Feature: show method for "connection refused" messages

OldhamMade:
This PR adds the method name to "connection refused" errors to aid debugging, as (for example) it's quite easy when duplicating a previous test as the basis for a new test to forget to change the method (e.g. `put` to `post`).


dcramer:
Thanks!

PR #55, opened 2015-04-07
Chain captured exceptions when exiting the context manager

ticosax:


responses.py
```diff
@@ -288,9 +290,15 @@ def unbound_on_send(adapter, request, *a, **kwargs):
     def stop(self):
         self._patcher.stop()
         if self.assert_all_requests_are_fired and self._urls:
-            raise AssertionError(
+            assertion_error = AssertionError(
                 'Not all requests has been executed {0!r}'.format(
                     [(url['method'], url['url']) for url in self._urls]))
+            if self.captured_exception:
+                try:
```
heyman: Doesn't this try-except block shadow still eat up other exceptions? Removing it, and instead just re-raising the captured exception, seems to give the correct behaviour for me. Am I missing something?

heyman:
Hey!

I've run into test failures that's hard to debug since responses is swallowing the real exceptions. I really appreciate the effort to resolve the issue, and I'm wondering why this pull request was closed?

responses.py
heyman: Ah, I now realised that the code is using exception chaining which is a Python 3 only feature. The result on python2 is that the original real exception will still go lost. On python 2, I think it would be better to just raise the captured exception.

PR #56, opened 2015-04-07
tox configuration file

ticosax:
@dcramer I do not know if you are interested by tox.
I just find it useful to deal with several python versions.

A next step would be to make travis relying on tox to reduce maintenance of 2 files.
I can do it if you like the idea.


dcramer:
I have dreams of a tox + travis.yml combination

Thanks!

Issue #62, opened 2015-05-19
Redirect option make the tests fail


elioscordo:
I am having this error now. Is this solved?
```
test_responses.py:299: in run
    resp = requests.get(url)
/Library/Python/2.7/site-packages/requests/api.py:55: in get
    return request('get', url, *_kwargs)
/Library/Python/2.7/site-packages/requests/api.py:44: in request
    return session.request(method=method, url=url, *_kwargs)
/Library/Python/2.7/site-packages/requests/sessions.py:361: in request
    resp = self.send(prep, *_send_kwargs)
responses.py:293: in unbound_on_send
    return self._on_request(session, requests, *a, *_kwargs)
responses.py:268: in _on_request
    if kwargs.get('allow_redirects') and response.is_redirect:
E   AttributeError: 'Response' object has no attribute 'is_redirect'
_____________________________________________________________ test_allow_redirects_samehost _____________________________________________________________
test_responses.py:369: in test_allow_redirects_samehost
    run()
test_responses.py:360: in run
    allow_redirects=True)
/Library/Python/2.7/site-packages/requests/api.py:55: in get
    return request('get', url, *_kwargs)
/Library/Python/2.7/site-packages/requests/api.py:44: in request
    return session.request(method=method, url=url, *_kwargs)
/Library/Python/2.7/site-packages/requests/sessions.py:361: in request
    resp = self.send(prep, *_send_kwargs)
responses.py:293: in unbound_on_send
    return self._on_request(session, requests, *a, *_kwargs)
responses.py:268: in _on_request
    if kwargs.get('allow_redirects') and response.is_redirect:
E   AttributeError: 'Response' object has no attribute 'is_redirect'
========================================================== 10 failed, 7 passed in 0.34 seconds ==========================================================
```

beliaev-maksim:
@markstory 
this one also could be closed as merged PR fixed the issue

PR #63, opened 2015-05-20
#62 is_redirect does not exist in requests 2.2 and earlier

hummus:
reproduced issue #62 on any requests <=2.2. 

It seems that _many_ more unit tests fail for requests <2.0, so I didn't even look at requests 0.x or 1.x ....


dcramer:
This looks good but we should avoid dynamic function binding and declare this at runtime

dcramer:
Thanks!

PR #70, opened 2015-06-09
Add Content-Length header

ferrix:
Web servers generally set this header and it is used to infer the length of the content still due to arrive. Makes sense to set it here too.


dcramer:
I'm not sure this is a safe default. Webservers don't always set the header as they dont always know the length. i.e. if you're streaming an iterator the length may not be known ahead of time

ferrix:
Yeah. I guess I can append the header when I really need it.

PR #73, opened 2015-07-10
Fix ConnectionError Emulation

fabian-gentner:
When raising a ConnectionError, the requests module sets the error's request attribute to the current request. The responses module should do the same.

Sample Script:
```
import requests
import responses

def req():
    try:
        requests.get('http://example.zalando.net')
    except Exception as e:
        return(e.request)
    else:
        assert False

print('Real:          ', req())
print('With responses:', responses.activate(req)())
```

Output:
```
Real:           <PreparedRequest [GET]>
With responses: None
````


dcramer:
Thanks!

PR #76, opened 2015-07-27
ENH: don't add default path if scheme is not http(s).

cournape:
Responses 0.4.0 and above change urls of the type "file://foo.bar" into "file://foo.bar/" for mocked urls. I don't think that's right.

This is a naive patch working around this, but I was not sure why http urls were changed to have a default path.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -101,11 +101,13 @@ class CallList(Sequence, Sized):
 
 
 def _ensure_url_default_path(url, match_querystring):
-    if _is_string(url) and url.count('/') == 2:
-        if match_querystring:
-            return url.replace('?', '/?', 1)
-        else:
-            return url + '/'
+    if _is_string(url):
+        p = urlparse(url)
+        if p.scheme in ("http", "https") and url.count('/') == 2:
+            if match_querystring:
+                return url.replace('?', '/?', 1)
+            else:
+                return url + '/'
     return url
 
 


dcramer:
I believe this was to ensure expressions would work correctly

i.e. http://foo.com *is* http://foo.com/ in HTTP (unless I'm mistaken, but Ive never seen otherwise), and our goal is to emulate the server's response

cournape:
In this case, do you agree we should whitelist http(s) ? Or should we instead handle `http://foo.com` and `http://foo.com/` equivalent when matching urls ?

I have some other PR following this one to fix handling of custom adapters

dcramer:
@cournape it might be safer matching, but maybe we should continue to whitelist since things like file: are different. Given that, I dont have a strong opinion on where it should go. If you're ok with it here I'll merge this.

cournape:
Looks fine as if for me.

dcramer:
One last thing. Could I get you to add a high level test for this? Can just prove that the behavior you saw broken now works. 

Thanks!

beliaev-maksim:
@markstory 
I do not think `requests` has an Adapter to handle `file` schema. That means that user must use custom adapter. Then, `responses` cannot inject own adapter to mock a response.
here is an example module how that should work: https://github.com/dashea/requests-file

to sum up, I do not think current PR will work. Until we see an example how to request a file without custom adapter. Propose to close it

markstory:
Closing as I don't think we need to handle this scenario with responses. Using an HTTP client to load a local file doesn't seem like something that requires mocking.

Issue #81, opened 2015-09-03
CHANGES needs to be updated for 0.5.0


singingwolfboy:

kdeldycke:
Seems to have been done in https://github.com/getsentry/responses/commit/03c00c9ff8676d240b71d3755d51cfc6e91341c7

PR #82, opened 2015-09-03
Allow passing a JSON body to `responses.add`

singingwolfboy:
It's a lot more convenient than writing a JSON-formatted string, and it automatically sets the Content-Type header for you!


dcramer:
this lgtm -- mind fixing the lint issues and ill merge?

singingwolfboy:
@dcramer Fixed!

Also, it would be good to make a new release with this functionality, and document it in the `CHANGES` file. Looks like that file could use a little love -- see #81.

dcramer:
@singingwolfboy we'll update the changes when we do a new release

PR #83, opened 2015-09-05
Indicate Python 3 support in setup.py

svisser:
This change indicates that the package supports Python 3. This ensures it'll pass automated checks such as https://caniusepython3.com/project/responses .

This would be very useful when automatically gathering all packages that do not yet support Python 3 in the projects that I'm working on.


graingert:
:+1:

PR #92, opened 2015-12-17
Don't swallow exceptions

### LINKED ISSUE #72: Responses eats all other Exceptions
Suppose I setup responses for a url and then a call a function where the url is being requests. If there is an exception in that URL, responses always throws an exception with AssertionError: Not all requests has been executed

This becomes very hard to debug when this isn't really the exception. Is there a better way to set this up?

heyman:
Currently, if an exceptions occurs within a RequestsMock with-block, it gets swallowed by the context manager. This fix makes the context manager instead re-raise the original exception.

These changes are based on #55. However, instead of chaining the exception with any assertion error due to all requests not being fired, I just re-raised the original exception since I think that makes more sense. However, if you prefer the exception-chaining behaviour, the tree at 4c00da8c4daab5d328a9b730433393971ac2a15e should do the work (#55 does still swallow exceptions on python 2 where exception chaining doesn't exist).

This should also fix #72.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -304,11 +304,16 @@ class RequestsMock(object):
             assertion_error = AssertionError(
                 'Not all requests have been executed {0!r}'.format(
                     [(url['method'], url['url']) for url in self._urls]))
-            if self.captured_exception:
+            if self.captured_exception[0]:
                 try:
                     six.reraise(*self.captured_exception)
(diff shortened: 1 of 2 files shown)


markstory:
Thanks for taking the time to put this together. This was fixed in 9d11c4d9

PR #39, opened 2014-10-15
Add feature to bypass responses per demand

ticosax:
This one is probably controversial.
`bypass` allows to bypass responses for selected requests.
Since `requests` is used in many libraries including elasticsearch clients for instance.
I would like in my test to *NOT* capture those calls in the context of `@responses.activate` that I'm using to capture calls to the web in the mean time.

Yep, I'm testing spaghetti code.

PR #93, opened 2015-12-17
Allow external requests

k4nar:
Sometimes,you want to mock a specific service, but you still want to be able to reach other services. With this PR, if `allow_external_requests` is `True`, calls to non mocked urls don't throw `ConnectionError`.

This is somewhat similar to #39.

EDIT: If someone is interested by this behavior, note that change may never be included in responses, as per @dcramer's [comment](https://github.com/getsentry/responses/pull/39#issuecomment-166695026) in #39.
If you need a way to mock an API while being able to make external requests, [request-mock](https://requests-mock.readthedocs.org/en/latest/) seems to be a good option.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -165,3 +165,30 @@ can be disabled by passing the ``assert_all_requests_are_fired`` value:
             rsps.add(responses.GET, 'http://twitter.com/api/1/foobar',
                      body='{}', status=200,
                      content_type='application/json')
+
+
+Allowing non mocked requests
+--------------------------------
+
+By default Responses will raise an exception if an url not mocked is called. This behavior can
+be changed by passing the ``allow_external_requests`` value:
+
+.. code-block:: python
+
+    import responses
+    import requests
+
+
+    def test_my_api():
+        with responses.RequestsMock(allow_external_requests=True) as rsps:
+            rsps.add(responses.GET, 'http://twitter.com/api/1/foobar',
+                     body='{}', status=200,
+                     content_type='application/json')
+
+            # use the mocked route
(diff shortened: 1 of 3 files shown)


bersace:
How to allow external requests with `responses.activate` ?

k4nar:
@bersace : Like with `assert_all_requests_are_fired`, you have to instanciate a `RequestsMock` objects. For example:

```python
rsps = RequestsMock(allow_external_requests=True)


@rsps.activate()
def test_some_stuff():
    ...
```

bersace:
ok. cool.

Daenyth:
Does this have a large benefit over doing the first request inside a `with` block and the second request outside it? It seems like it would make it easy to make mistakes when writing a test with this enabled.

k4nar:
@Daenyth : If you can make your request outside the with then no, there is no benefit.
However if you are testing a function calling another service that you don't want to mock then you don't really have the choice.

adamtheturtle:
To add my 👍 to this: I wanted to add tests to a legacy function which uses `requests` to interact with two services. In order to use `responses` for one service, I'd have to create mocks for both, which I'm just not ready to do. Fortunately, switching to `requests-mock` wasn't too painful but I'd have rather not done this.

dcramer:
i realize this is ancient, but I do want to get support for this in. That said, I'd like to have the behavior be something like ``responses.allow_external_domains('domain.com', 'otherdomain.com')``, since one of responses goals is to ensure you dont accidentally rely on being network connected

kathawala:
I think this PR should be closed, as the functionality proposed here was added by #174

k4nar:
Cool thanks 👍 .

Issue #79, opened 2015-08-18
Feature request: easy assertions on post data


alexpdp7:
``` python
@responses.activate
def foo():
        responses.add(responses.POST, 'http://foo.bar')
        requests.post('http://foo.bar', {'qux': 'quux'})
```

I would like to assert that I've posted qux=quux, but doing this completely well is uncomfortable (esp. if you want to be py2-3 compatible).

Ideally, I would like to do:

``` python
assert responses.calls[0].data = {'qux': 'quux'}
```

or something equivalent.

danield-hyprbrands:
+1

greenmoon55:
+1

alexntt:
Any chance of merging @alexanderyoung's PR? I've come across another instance where I wish this was done

berinhard:
Same here!

rdrey:
Would be nice if it could "match" on body params. That way it would raise if different params were sent (nothing matched).

dcramer:
This is possible today, and while we could add a million new APIs, I'm closing this to avoid noise.

```
    assert responses.calls[0].response.text == '{"value": 6}'
    assert (
        responses.calls[0].response.headers['request-id'] ==
        '728d329e-0e86-11e4-a748-0c84dc037c13'
    )
```

PR #94, opened 2015-12-22
Match request body

alexanderyoung:
Use case: Handle different POST events.
May address https://github.com/getsentry/responses/issues/79

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -130,7 +130,8 @@ class RequestsMock(object):
 
     def add(self, method, url, body='', match_querystring=False,
             status=200, adding_headers=None, stream=False,
-            content_type='text/plain', json=None):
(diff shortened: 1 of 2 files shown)


alexanderyoung:
@dcramer

dcramer:
I think, if I were to spend any time expanding the API (which is already getting bloated), I'd rather we just add some kind of side_effect/callback format to be used for anything beyond simple cases.

```
handler(request): if request.headers['Foo-Bar'] == 'baz': return response; return 500Response

responses.add(url, method='POST', handler)
```

markstory:
Thanks for taking the time to put this together, but in the intervening years another approach has been merged and responses now has [request parameter matching](https://github.com/getsentry/responses#matching-request-parameters)

Issue #95, opened 2016-01-13
reponses requires coverage < 4.0.0 but many systems ships coverage >= 4.0.0


Germano0:
I am following a procedure that will bring responses inside official Fedora repositories.
[responses][1] needs coverage < 4.0.0 but Fedora 23 ships coverage 4.0.3, so during RPM package creation, when the system runs checks/tests I obtain errors
```bash
Traceback (most recent call last):
  File "setup.py", line 94, in <module>
    'Topic :: Software Development'
  File "/usr/lib64/python3.4/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/usr/lib64/python3.4/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/usr/lib64/python3.4/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/usr/lib/python3.4/site-packages/setuptools/command/test.py", line 134, in run
    self.distribution.fetch_build_eggs(self.distribution.tests_require)
  File "/usr/lib/python3.4/site-packages/setuptools/dist.py", line 313, in fetch_build_eggs
    replace_conflicting=True,
  File "/usr/lib/python3.4/site-packages/pkg_resources/__init__.py", line 836, in resolve
    dist = best[req.key] = env.best_match(req, ws, installer)
  File "/usr/lib/python3.4/site-packages/pkg_resources/__init__.py", line 1074, in best_match
    dist = working_set.find(req)
  File "/usr/lib/python3.4/site-packages/pkg_resources/__init__.py", line 711, in find
    raise VersionConflict(dist, req)
pkg_resources.VersionConflict: (coverage 4.0.3 (/usr/lib64/python3.4/site-packages), Requirement.parse('coverage<4.0.0,>=3.7.1'))
```

Is there anything we can do?

[1]: https://github.com/getsentry/responses/blob/master/setup.py#L33

dcramer:
Personally I strongly disagree with packaging Python libraries with distros. The dependency hell that I've seen requests go through, along with the obvious fact that it simply doesnt work due to Python not supporting multiple lib versions, makes me very much against the pattern.

With that said, if you want to put the effort into ensuring it works with both coverage 3.x as well as 4.x, I'm not against bumping the requirements here.

AdamWill:
[...]

PR #96, opened 2016-01-22
stop pinning coverage requirement to < 4.0.0 (#95)

AdamWill:
Coverage testing seems to work fine at least with 4.1b1 and
Python 3.5 (tested on Fedora Rawhide), and this kind of pinning
is not viable in the long term and causes problems for
downstream distributors who include the newer coverage.


AdamWill:
The failure doesn't look to have anything to do with this change, rather travis' attempt to install requests failed...

dcramer:
thanks!

PR #100, opened 2016-01-28
Allow empty list/dict as json object

elatomo:
Currently, is not possible to pass an object that evaluates to `False` as a `json` argument:

```python
    responses.add(responses.GET, url, json=[])  # or `json={}`
    req = requests.get(url)
    assert req.headers['Content-Type'] == 'application/json'  # fails!


dcramer:
thanks!

PR #101, opened 2016-02-05
Add multi-response support

MichaelAquilina:
I often find myself needing to test something like the following:

`Test that if a response fails the first time, my code is able to retry the request a certain amount of times before giving up.`

Currently (as far as I could tell from the documentation and reading the code at first glance) this didnt seen possible and I usually resorted to using mock instead.

Here is a (very simplified) example of a test I have written before to test a scenario where my code should retry once before giving up.

```
    @mock.patch('requests')
    def test_retry_on_error(self, requests):
        response = mock.Mock()
        requests.get.return_value = response
        response.iter_lines.side_effect = [requests_exceptions.HTTPError(), []]
        my_requesting_code()
```

It would be more ideal to use responses because that's more readable.

While trying to add this feature to the responses library I noticed that you are essentially already supporting this use case through the `assert_all_requests_are_fired` option which removes urls from `self._urls` as they are used. 

However this is very non-obvious and required me to actually delve into the code to figure this out!

This pull request therefore adds a more obvious "multi_response" flag as a parameter which essentially enables the same functionality but without asserting all responses were fired at the end. 

I wrote a test for the multi response case and updated `test_assert_all_requests_are_fired` to ensure the code still works as expected.

I'm open to updating the code as you see fit as I understand the way I use the constructor parameters is a bit hacky but I think a feature like this could be very valuable for users.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -119,11 +119,11 @@ class RequestsMock(object):
     POST = 'POST'
     PUT = 'PUT'
 
-    def __init__(self, multi_response=True, assert_all_requests_are_fired=True):
+    def __init__(self, multi_response=False, assert_all_requests_are_fired=True):
         self._calls = CallList()
         self.reset()
-        self.assert_all_requests_are_fired = multi_response
-        self.multi_response = multi_response
+        self.assert_all_requests_are_fired = assert_all_requests_are_fired
+        self.multi_response = assert_all_requests_are_fired or multi_response
 
     def reset(self):
         self._urls = []
--- a/responses.py
+++ b/responses.py
@@ -119,7 +119,11 @@ class RequestsMock(object):
     POST = 'POST'
     PUT = 'PUT'
 
-    def __init__(self, multi_response=False, assert_all_requests_are_fired=True):
+    def __init__(
+        self,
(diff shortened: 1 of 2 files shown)


dcramer:
I thought I wrote some notes about what I thought would be a good API, but can't find the ticket.

I'd prefer something of an overload approach that keeps the API simple (accepting PRs has taken what once we a simple project and made it complex/inconsistent).

```
response_list = [
    Response(json='...', status=201),
    Response(json='...', status=200),
]
responses.add(responses.GET, 'http://example.com', response_list)
```

Now the complication would be around "what happens when the iterator exhausts? Should it start over? Should it repeat the last request? Should it error? I think we could say it should error (unregister a response after its executed).

I think we really should revisit the responses.add API, and I simply don't have time to sit down and do it lately.

adamchainz:
> Now the complication would be around "what happens when the iterator exhausts? Should it start over? Should it repeat the last request? Should it error? I think we could say it should error (unregister a response after its executed)

I agree with erroring. One can always make it cycle with `itertools.cycle` or pass in a generator for more nuanced behaviour.

Issue #102, opened 2016-02-07
Documentation misleading regarding default behaviour of assert_all_requests_are_fired


inglesp:
The README says "By default Responses will raise an assertion error if a url was registered but not accessed".  However, this is not true if responses is not used as a context manager.  See https://github.com/getsentry/responses/blob/master/responses.py#L308.

PR #103, opened 2016-02-07
Clarify docs about assert_all_requests_are_fired

inglesp:
Addresses #102


dcramer:
Thanks! Not super happy about the differences between context manager and decorator (this shouldn't be the case), but it's better to be correct here.

PR #105, opened 2016-03-11
Return URL or Callback for Quick Updating

jpoehnelt:
Sometimes it may be nice to quickly mock multiple responses within a single test with only minor changes in the response. This allows that to happen.

```
fake = responses.add(responses.GET, url, body="something", status=200)
r = requests.get(url)
assert r.status_code == fake['status']

fake['status'] = 400
r = requests.get(url)
assert r.status_code == fake['status'] # 400

fake['body'] = "something else"
r = requests.get(url)
assert r.text == fake['body'] # "something else"
```

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -155,6 +155,7 @@ class RequestsMock(object):
             'adding_headers': adding_headers,
             'stream': stream,
         })
+        return self._urls[-1]
 
     def add_callback(self, method, url, callback, match_querystring=False,
                      content_type='text/plain'):
@@ -168,6 +169,7 @@ class RequestsMock(object):
             'content_type': content_type,
             'match_querystring': match_querystring,
         })
+        return self._urls[-1]
 
     @property
     def calls(self):


beliaev-maksim:
@markstory 
I am not confident that we want to promote such functionality. 
I would rather create a `Response` object and add it multiple times, modifying attributes each time
https://github.com/getsentry/responses/blob/f6c71ae70aa1645f6fdda4eb236042109c225e2e/responses/test_responses.py#L1934


your thoughts?

markstory:
>  I would rather create a Response object and add it multiple times, modifying attributes each time

I agree. Having mutable objects come out of responses increases the chance for temporal coupling in tests. I also prefer the flow where responses are created in the test, and then attached to response mocks.

beliaev-maksim:
@markstory
OK, then propose to close the PR

markstory:
Thanks for taking the time to put this together @jpoehnelt. I don't want to encourage user code to mutating the response objects. If you need this behavior you can create your own response instances and attach them to responses similar to the test case example linked to earlier in the issue.

PR #107, opened 2016-03-31
support for providing query params as a dict in responses.add

abehrens:
I wanted to add a mechanism to have the query string built in responses instead of doing so by hand.  I stole some structure from the requests library and tossed in a test and a snippet in the readme.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -300,7 +300,8 @@ class RequestsMock(object):
                 if v is not None:
                     result.append(
                         (k.encode('utf-8') if isinstance(k, str) else k,
-                        v.encode('utf-8') if isinstance(v, str) else v))
+                            v.encode('utf-8') if isinstance(v, str) else v)
+                    )
         return urlencode(result, doseq=True)
 
     def start(self):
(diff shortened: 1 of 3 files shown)


danihodovic:
@abehrens are there any future plans on integrating this? It would be a very useful feature.

AbdealiJK:
Came here looking for the same functionality - any plans of adding it ?

responses.py
```diff
@@ -130,14 +131,20 @@ def reset(self):
...
+            content_type='text/plain', json=None, query_params=None):
 
         # if we were passed a `json` argument,
         # override the body and content_type
         if json is not None:
             body = json_module.dumps(json)
             content_type = 'application/json'
 
+        # if we are passed a query_params dict, create the query
+        # string and append to the url
```
merwok, line 143: Should the presence of this parameter imply match_querystring=True ?

beliaev-maksim:
@markstory 
I think this may be closed since idea is the same as in #397

merwok:
The other PR does provide a way to do the same thing, although this one was easier to use.

beliaev-maksim:
@merwok 
you can find comments in #397 to see why was decided to go via matchers.

PR #108, opened 2016-04-11
Allow the fragment identifier to be ignored when matching urls.

zamzterz:
Defaults to not forcing match of the fragment identifier, similar to the handling of querystrings.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -20,9 +20,9 @@ except ImportError:
     from urllib3.response import HTTPResponse
 
 if six.PY2:
-    from urlparse import urlparse, parse_qsl
+    from urlparse import urlparse, parse_qsl, urlunparse
 else:
-    from urllib.parse import urlparse, parse_qsl
+    from urllib.parse import urlparse, parse_qsl, urlunparse
 
 if six.PY2:
     try:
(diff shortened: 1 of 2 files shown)


dcramer:
I'm not against the idea of this change, but I dont know if "fargment identifier" is really a common term. This is treating a fragment as if its a query string (which I realize, once upon a time, was common). For things like this I definitely would prefer we just create a better way of handling requests so you can specify a callback to deal w/ validation.

rebeckag:
"Fragment" is probably the most common term, but since it's "querystring" (instead of just "query") I just chose "fragment identifier" [based on Wikipedia](https://en.wikipedia.org/wiki/Fragment_identifier). I have no issues with changing the naming to either "fragment" or something else if you have any suggestions.

I agree with your idea about a more general way of delegating the request validation, but my use case is actually to be able to ignore the fragment. And since that functionality is provided for query strings it would be nice, for consistency, if the same behaviour is implemented for fragments by default (without needing to provide my own custom validation callback which just ignores the fragment, but needs to otherwise match the url).

beliaev-maksim:
@markstory 
please close this PR

markstory:
Closing as fragment matching has been added as part of the matchers module.

PR #109, opened 2016-04-19
Make the URL parameter to add() and add_callback() interpreted similarly (fixes #32)

### LINKED ISSUE #32: Inside vagrant vm... "Connection refused"
I'm getting "Connection refused" error from requests, and this seems to prevent responses from returning its fake response.

I have:

```
    responses.add_callback(responses.GET, 'http://google.com',
                           callback=mycallback)
```

If I inspect `responses.calls._calls` I can see:

`Call(request=<PreparedRequest [GET]>, response=ConnectionError(u'Connection refused: http://google.com/',))`

so responses is activated ok and has caught the request

okomestudio:
Uncommenting out the line appears to fully fix the open issue:

https://github.com/getsentry/responses/issues/32

The current code (with comment-out) makes `add()` and `add_callback()` change behavior with regard to input `url` parameter, which does not make sense.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -159,7 +159,7 @@ class RequestsMock(object):
     def add_callback(self, method, url, callback, match_querystring=False,
                      content_type='text/plain'):
         # ensure the url has a default path set if the url is a string
-        # url = _ensure_url_default_path(url, match_querystring)
+        url = _ensure_url_default_path(url, match_querystring)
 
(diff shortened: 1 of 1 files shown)


okomestudio:
I am closing this since I think that the bug has been fixed with all the changes that have happened since the PR was opened.

PR #110, opened 2016-04-19
Properly retain session cookies (resolves #80)

### LINKED ISSUE #80: Sessions lose cookies
Hi, 

It seems that there's a problem with the mocked HTTP requests returned when using a session - the session cookie is returned in the response object, but does not get set in the session.

I am pretty sure it's down to the extract_cookies_to_jar function in cookielib/cookies.py:

``` python
def extract_cookies_to_jar(jar, request, response):
    """Extract the cookies from the response into a CookieJar.

    :param jar: cookielib.CookieJar (not necessarily a RequestsCookieJar)
    :param request: our own requests.Request object
    :param response: urllib3.HTTPResponse object
    """
    if not (hasattr(response, '_original_response') and
            response._original_response):
        return
    # the _original_response field is the wrapped httplib.HTTPResponse object,
    req = MockRequest(request)
    # pull out the HTTPMessage with the headers and put it in the mock:
    res = MockResponse(response._original_response.msg)
    jar.extract_cookies(res, req)
```

The _original_response attribute is missing from the mocked HTTPResponse, so no cookies get pulled out into the jar. 
I'm trying to come up with a solution and if I do I'll post here, but I should thought I'd let the community at large know. Perhaps I'm doing something silly :-)

okomestudio:
This resolves https://github.com/getsentry/responses/issues/80.

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/responses/110)

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -20,6 +20,11 @@ except ImportError:
     from urllib3.response import HTTPResponse
 
 if six.PY2:
+    from requests.packages.urllib3._collections import HTTPHeaderDict
+else:
+    from http.client import HTTPMessage
+
+if six.PY2:
     from urlparse import urlparse, parse_qsl
 else:
     from urllib.parse import urlparse, parse_qsl
@@ -40,6 +45,33 @@ def wrapper%(signature)s:
     with responses:
         return func%(funcargs)s
 """
+
+
+class _OriginalResponse(object):
+    def __init__(self, headers):
+        if six.PY2:
+            self.msg = HTTPHeaderDict()
+            if 'set-cookie' in headers:
+                cookies = Cookies.from_request(headers.pop('set-cookie'))
+                for _, v in cookies.items():
+                    self.msg.add('set-cookie', '='.join([v.name, v.value]))
+            for k, v in headers.items():
+                self.msg[k] = v
+        else:
+            self.msg = HTTPMessage()
+            if 'set-cookie' in headers:
+                cookies = Cookies.from_request(headers.pop('set-cookie'))
+                for _, v in cookies.items():
+                    self.msg.add_header('set-cookie',
+                                        '='.join([v.name, v.value]))
+            for k, v in headers.items():
+                self.msg[k] = v
+
+    def close(self):
+        pass
+
+    def isclosed(self):
+        return True
 
 
 def _is_string(s):
@@ -264,6 +296,7 @@ class RequestsMock(object):
             body=body,
             headers=headers,
             preload_content=False,
+            original_response=_OriginalResponse(headers)
         )
 
         response = adapter.build_response(request, response)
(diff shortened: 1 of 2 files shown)


dcramer:
Looks like this is broken on requests 2.0

okomestudio:
Now barfing on Python 2.6... :/

madssj:
Any reason not to merge this?

FelixSchwarz:
@madssj Basically the project lacks a maintainer which can drive the project forward without breaking old code needlessly. https://github.com/getsentry/responses/pull/123#issuecomment-253850645 has more details.

andi-bigswitch:
@dcramer any chance of getting this PR merged? #80 seems like a pretty clear bug to me (`responses` behaves differently from `requests` in that session objects do not retain cookies).

Depending on whether @okomestudio is still available to address comments / fix merge conflicts, I'd also be happy to help to get this merged.

okomestudio:
I could take a look at this again and make changes if necessary. However the issues (on requests and Python 2.6 explicitly mentioned in the thread) should have been addressed already given the tests are passing. It's been a while since the PR was created and probably needs to be checked against the current master though.

okomestudio:
The merge conflicts have been resolved. Please go ahead with review and such as appropriate.

responses.py
```diff
@@ -402,9 +408,42 @@ def get_response(self, request):
...
+        headers = headers.copy()
+
+        if "set-cookie" in headers:
+            cookiejar = _cookies_from_headers(headers)
+            headers.pop("set-cookie", None)
+        else:
+            cookiejar = None
+
+        if six.PY2:
+            items = [k + ": " + v for k, v in headers.iteritems()]
```
markstory: This looks like it could hit a UnicodeError if the header contains a non-ascii byte.

okomestudio: Could you first take a look at #269, which I just created? I was adding a couple more tests on headers/cookie handling and that would help me better understand how to go about handling this issue.

test_responses.py
```diff
@@ -664,6 +664,34 @@ def run():
...
+    @responses.activate
+    def run():
+        responses.add_callback(responses.GET, url, request_callback)
+        session = requests.Session()
+        resp = session.get(url)
+        assert resp.text == "test callback"
+        assert resp.status_code == status
+        assert "session_id" in resp.cookies
+        assert resp.cookies["session_id"] == "12345"
+        assert resp.cookies["a"] == "b"
```
retracile: A `Set-Cookie: session_id=12345; a=b; c=d` header should create a `session_id` cookie with `a` and `b` attributes; it should not create `a` and `b` cookies.  See PR#285 for more detail and an alternative solution.

okomestudio: Thanks for catching the issue. I'm all for your PR to be merged instead of this one so long as it resolves the long-standing issue.

PR #112, opened 2016-04-28
Add section for match_querystring arg to README

steven-mercatante:
After some head scratching and source code surfing, I found the `match_querystring` argument. It'd be nice if this was in the docs :)

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/responses/112)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -166,3 +166,27 @@ the ``assert_all_requests_are_fired`` value:
             rsps.add(responses.GET, 'http://twitter.com/api/1/foobar',
                      body='{}', status=200,
                      content_type='application/json')
+
+
+Matching query strings
+----------------------
+
+When mocking a URL that contains a query string, pass `match_querystring=True` to `responses.add()`.
+
+.. code-block:: python
+
+    import responses
+    import requests
+
+    @responses.activate
+    def test_my_api():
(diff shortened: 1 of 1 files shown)


whyscream:
+1, not finding this cost me too much time today :/

README.rst
```diff
@@ -166,3 +166,27 @@ the ``assert_all_requests_are_fired`` value:
...
+    def test_my_api():
+        responses.add(responses.GET, 'http://twitter.com/api/1/foobar?baz=true',
+                      json={"error": "not found"}, status=404, match_querystring=True)
+
+        resp = requests.get('http://twitter.com/api/1/foobar?baz=true')
+
+        assert resp.json() == {"error": "not found"}
+
+        assert len(responses.calls) == 1
+        assert responses.calls[0].request.url == 'http://twitter.com/api/1/foobar'
```
sliwinski-milosz: Sorry but your test will actually fail as your url is missing query string:
```python
responses.calls[0].request.url == 'http://twitter.com/api/1/foobar?baz=true'
```

markstory:
Thank you for taking the time to put this together. The docs currently have a section on the `match_querystring` parameter and I don't think a sample is necessary given the current content.

PR #114, opened 2016-05-30
Add headers matching support

### LINKED ISSUE #68: headers parameter for responses.add
Would it be possible for `responses.add` to support a `headers` parameter? Having to create a callback just to add headers seems a bit more work than it should be.

a-pertsev:
Sometimes remote server response depends on headers were sent to, so we need to have some opportunity to mock different responses basing on request headers.

Example:
```
import requests

SOME_URL = 'http://example.com/some/url'
response1 = requests.get(SOME_URL, headers={'Content-Type': 'application/json'}
response2 = requests.get(SOME_URL, headers={'Content-Type': 'application/xml'}
```
 
and in tests we need to have:
```
 responses.add(responses.GET, SOME_URL, headers={'Content-Type': 'application/json'}, body=some_json_body)
 responses.add(responses.GET, SOME_URL, headers={'Content-Type': 'application/xml'}, body=some_xml_body)
```

@dcramer will be happy to get feedback from you.

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/responses/114)

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -130,7 +130,7 @@ class RequestsMock(object):
 
     def add(self, method, url, body='', match_querystring=False,
             status=200, adding_headers=None, stream=False,
-            content_type='text/plain', json=None):
+            content_type='text/plain', json=None, headers=None):
 
         # if we were passed a `json` argument,
         # override the body and content_type
@@ -149,6 +149,7 @@ class RequestsMock(object):
             'url': url,
             'method': method,
             'body': body,
+            'headers': headers,
             'content_type': content_type,
             'match_querystring': match_querystring,
             'status': status,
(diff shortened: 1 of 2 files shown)


prokaktus:
Just interesting, why do you decide, that response headers should match to request headers? I have many examples, where I expect specific response header, but don't send it in request.

a-pertsev:
@prokaktus well
1. responses already has "adding_headers" parameter to pass it to HTTPResponse:
https://github.com/a-pertsev/responses/blob/master/responses.py#L258
2. responses.add method first arguments "method" and "url" are both used to match requests so i added "headers" parameter to keep it consistent. 
But mixed response and request parameters in responses.add function is really confusing, you are right =/

a-pertsev:
@dcramer could you tell please if i should resolve conflicts or close this PR?

dcramer:
If you’d be willing to rebase this I think we can get it merged in

a-pertsev:
@dcramer i see no elegant way to implement this feature in requests project after this changes
https://github.com/a-pertsev/responses/commit/24859ba783a6e10735eab1da36ff0982e964aae4
so i think i should apologize for being so annoying, i will close this PR and will implement this features in my own repo

PR #115, opened 2016-06-21
Provide support for timeout

bmwant:
It's useful to have an ability of specifying the delay when making a request. It helps in integration tests and allows imitating network latency.
Also `timeout` for request itself allow to test for `ConnectionTimeout` errors that is another one case to check in unittests.
## Changes
- Add makefile target for only launching py.test in verbose mode
- Update `.gitignore` file
- Add `timeout` keyword for `add` and `add_callback` methods

⌛ timeout ⏳

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -12,8 +12,13 @@ from collections import namedtuple, Sequence, Sized
 from functools import update_wrapper
 from cookies import Cookies
 from requests.utils import cookiejar_from_dict
-from requests.exceptions import ConnectionError, ConnectTimeout
+from requests.exceptions import ConnectionError
 from requests.sessions import REDIRECT_STATI
+
+try:
+    from requests.exceptions import ConnectTimeout
+except ImportError:
+    from requests.exceptions import Timeout as ConnectTimeout
 
 try:
     from requests.packages.urllib3.response import HTTPResponse
--- a/responses.py
+++ b/responses.py
@@ -2,6 +2,7 @@ from __future__ import (
     absolute_import, print_function, division, unicode_literals
 )
 
+import time
 import inspect
 import json as json_module
 import re
@@ -11,7 +12,7 @@ from collections import namedtuple, Sequence, Sized
 from functools import update_wrapper
 from cookies import Cookies
 from requests.utils import cookiejar_from_dict
-from requests.exceptions import ConnectionError
+from requests.exceptions import ConnectionError, ConnectTimeout
 from requests.sessions import REDIRECT_STATI
 
 try:
@@ -130,7 +131,7 @@ class RequestsMock(object):
 
     def add(self, method, url, body='', match_querystring=False,
             status=200, adding_headers=None, stream=False,
-            content_type='text/plain', json=None):
+            content_type='text/plain', json=None, timeout=None):
 
         # if we were passed a `json` argument,
         # override the body and content_type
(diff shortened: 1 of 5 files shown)


responses.py
```diff
@@ -282,6 +290,15 @@ def _on_request(self, adapter, request, **kwargs):
...
         self._calls.add(request, response)
 
+        request_timeout = kwargs.get('timeout')
+        response_timeout = match.get('timeout')
+        if request_timeout is not None and response_timeout is not None:
+            if request_timeout < response_timeout:
+                raise ConnectTimeout()
+
+        if response_timeout:
+            time.sleep(response_timeout)
```
hansmosh: I get what you're going for but unit tests should be fast (very fast) and adding sleeps is generally not a good idea.

test_responses.py
```diff
@@ -390,3 +396,56 @@ def run():
...
+
+
+def test_request_timeout_expired():
+    url = 'http://example.com/'
+    response_timeout = 0.2
+    request_timeout = 0.1
+
+    @responses.activate
+    def run():
+        responses.add(responses.GET, url, timeout=response_timeout)
```
hansmosh: You could get the same result by using `body=ConnectTimeout` here. No need for a new `timeout` argument. https://github.com/getsentry/responses/blob/master/responses.py#L243

test_responses.py
```diff
@@ -390,3 +396,56 @@ def run():
...
+        with pytest.raises(ConnectTimeout):
+            requests.get(url, timeout=request_timeout)
```
hansmosh: What I do think would be a good feature would be to assert that timeout was used. Something like `self.assertEqual(responses.calls[0].request.timeout, request_timeout)`

test_responses.py
amfarrell: Actually, if you do that, then within the function you linked 
```
    def _on_request(self, adapter, request, **kwargs):
        match = self._find_match(request)
        # TODO(dcramer): find the correct class for this
        if match is None:
            error_msg = 'Connection refused: {0} {1}'.format(request.method,
                                                             request.url)
            response = ConnectionError(error_msg)
            response.request = request

            raise response

        if 'body' in match and isinstance(match['body'], Exception):
            self._calls.add(request, match['body'])
            raise match['body']
```
`match['body']` will equal `ConnectTimeout` and therefore `isinstance(match['body'], Exception)` will be False.

You'll want to use `body=ConnectTimeout()` so that you actually pass an instance.

responses.py
bmwant: agree, will rewrite it using different approach

bmwant:
@hansmosh @amfarrell I have added proper testing for the timeout.
Regarding to the extra keyword argument `timeout` I didn't get what you are trying change here. 
I want fine grained control over the calls and I probably want to provide `body` parameter with the `timeout` parameter at the same time. Without explicitly specifying as function's argument that would not be possible. Could you explain in more details why this approach should be changed?

bmwant:
seems like nobody wants this in, closing.

aoberoi:
i'm also interested in this feature. i'm having difficulty testing my library's API that accepts a `timeout` argument and verifying that the `ConnectTimeout` error is actually raised.

i explored using `add_callback()` and within my callback using `time.sleep()` (it is acceptable for me to add a 0.05s latency to this test), but it does not trigger the `ConnectTimeout`.

when i use my library without `responses`, the `ConnectTimeout` is raised as expected. therefore it seems `responses` is interfering with (or not accurately simulating) the behavior.

is there anything i can do to help add this feature?

bmwant:
@aoberoi reopen PR and rebase it on latest master; make sure tests pass
probably that might help moving this forward

PR #117, opened 2016-06-27
Added support for removing URLs from responses

rafaelsierra:
Hi Guys, I want to suggest the addition of a `remove` method to allow removing URL/method pairs from being mocked.

It happens quite often for me that I need to change a response for the same URL more than once during tests (because the same URL might return different kind of content depending on the input data).

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/responses/117)

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -156,6 +156,15 @@ class RequestsMock(object):
             'stream': stream,
         })
 
+    def remove(self, method, url, all_instances=True):
+        for mocked_url in self._urls:
+            if mocked_url['method'] == method and mocked_url['url'] == url:
+                self._urls.remove(mocked_url)
+                # responses allows for multiple registrations of the same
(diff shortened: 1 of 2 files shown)


joscarsson:
Is this perhaps solving the same use-case as my PR https://github.com/getsentry/responses/pull/122? Would probably be good to have both `replace` and `remove` though.

rafaelsierra:
@joscarsson yes, I agree that replacing is also needed (it was one of the reasons I made the remove in the first place).

I would like to have both also

joscarsson:
@rafaelsierra I added the `remove` functionality to #122 too, hope you don't mind. Was able to reuse a little code.

rafaelsierra:
Sure, no problem.

Closing in favor of #122

PR #118, opened 2016-06-29
Fixes #47: document an example pytest fixture

### LINKED ISSUE #47: Provide a py.test fixture in addition to the decorator
When using py.test for tests it's usually nicer to activate features like this using a fixture. The usage with the fixture (assuming it's named `responses`) could be like this:

``` python
def test_my_api(responses):
    responses.add(responses.GET, 'http://twitter.com/api/1/foobar',
                  body='{"error": "not found"}', status=404,
                  content_type='application/json')

    # ...
```

I.e. instead of the decorator you'd use a `responses` fixture and instead of calling `add` as a global from the imported module (if you register it as a pytest plugin the user doesn't need to import it at all) you'd call it on the object returned by the fixture function (which could simple be the module `responses`, but a wrapper would be cleaner of course).

tomviner:
Fixes https://github.com/getsentry/responses/issues/47

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/responses/118)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,7 +2,7 @@ Responses
 =========
 
 .. image:: https://travis-ci.org/getsentry/responses.png?branch=master
-	:target: https://travis-ci.org/getsentry/responses
+    :target: https://travis-ci.org/getsentry/responses
 
 A utility library for mocking out the `requests` Python library.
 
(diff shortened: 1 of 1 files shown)


bmwant:
@tomviner Could you also add an example for usage of such a fixture? Add update indent for decorator, please, there is an extra space.

tomviner:
Thanks for review @bmwant, I've added an example usage.

latusaki:
This PR should be closed now?

README.rst
```diff
@@ -147,6 +147,23 @@ Responses as a context manager
         resp = requests.get('http://twitter.com/api/1/foobar')
         resp.status_code == 404
 
+Responses as a pytest fixture
+-----------------------------
+
+.. code-block:: python
+
+    @pytest.yield_fixture
```
markstory: Shouldn't this be `pytest.fixture`?

bmwant: In newer versions - yes, but that's ok to have it in example, we do not pin to specific pytest version here

markstory:
Rebased and merged in 3bc26d57bfbf02a570b6df30212a56f183349e06

PR #122, opened 2016-08-11
Add support for removing and replacing existing mocked URLs

joscarsson:
This PR adds support for `responses.replace()`. The problem it is solving is the following: we usually have a base fixture of mocked API calls which we use like below in tests.

``` python
@pytest.mark.usefixtures('mock_api_calls')
def some_test():
   pass
```

The fixture itself is defined as:

``` python
@pytest.fixture()
def mock_api_calls():
    responses.add(responses.GET, 'http://someurl.com/one')
    responses.add(responses.GET, 'http://someurl.com/two')
    responses.add(responses.GET, 'http://someurl.com/three')
    responses.add(responses.GET, 'http://someurl.com/four')
```

However, in the specific test we then want to change **one** of the API calls to return a different response. This wasn't possible previously, as `responses.add()` always appends to `_urls`. With `replace()`, we can now do:

``` python
@pytest.mark.usefixtures('mock_api_calls')
def some_test():
  responses.replace(responses.GET, 'http://someurl.com/two', body='different')
```

Implementation-wise I broke out the creation of the entry into `_build_entry()` to be able to reuse it from `replace()`. I also reused `_find_match()` by adding an extra keyword for `exact_match` (as I figured we don't want to replace based on regex).

Thoughts?

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/responses/122)

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -134,6 +134,26 @@ class BaseResponse(object):
         self.url = _ensure_url_default_path(url)
         self.call_count = 0
 
+    def __eq__(self, other):
+        if not isinstance(other, BaseResponse):
+            return False
+
+        if self.method != other.method:
+            return False
+
+        # Can't simply do a equality check on the objects directly here since __eq__ isn't
+        # implemented for regex. It might seem to work as regex is using a cache to return
+        # the same regex instances, but it doesn't in all cases.
+        self_url = self.url.pattern if isinstance(
+            self.url, re._pattern_type) else self.url
+        other_url = other.url.pattern if isinstance(
+            other.url, re._pattern_type) else other.url
+
+        return self_url == other_url
+
+    def __ne__(self, other):
+        return not self.__eq__(other)
+
     def _url_matches_strict(self, url, other):
         url_parsed = urlparse(url)
         other_parsed = urlparse(other)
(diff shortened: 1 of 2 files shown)


joscarsson:
[Linting fails on Python 2.6](https://travis-ci.org/getsentry/responses/jobs/151570277), seems `flake8` is trying to use `OrderedDict`?

joscarsson:
I extended this PR to also include support for removing a mocked URL (a bit weird to have support for `replace` but not `remove`, and I was able to reuse a little code - there's [another PR](https://github.com/getsentry/responses/issues/117) for implementing only the `remove` support if you'd rather consider that though).

sgillies:
@joscarsson I've proposed a fix for linting in #127.

joscarsson:
@sgillies: Nice, when that's merged into `master` I can rebase off of that then :)
@dcramer: What do you think about this PR? Any chance it might be merged anytime soon?

dcramer:
@joscarsson https://github.com/getsentry/responses/pull/123#issuecomment-253850645

I'd rather give more people access to maintain the project, as I can't invest the time to QA and ensure compatible with Sentry's codebase on many of these changes.

joscarsson:
@dcramer: I understand, thanks for getting back to me. This PR is fully backwards-compatible though, and I assume it would be released under a new version so I don't see how it would affect Sentry's codebase.

Perhaps add a warning to the README that PRs aren't currently accepted and that you're looking for maintainers? A little frustrating spending time writing PRs that won't be merged/considered.

ownaginatious:
@joscarsson looks like a new version just got released to PyPI. Could you resolve the conflicts in this PR? Maybe the maintainers will merge it now :)

joscarsson:
@ownaginatious: Thanks for the ping, I've resolved the conflicts and updated the PR. 

@dcramer: if you have time I'd appreciate it if you took a look (or close the PR if you don't want the feature). Travis build still fails for Python 2.6, but that seems to be because a lib have dropped support for Python 2.6 and not because of the changes in this PR.

responses.py
```diff
@@ -332,10 +339,34 @@ def add(self, method_or_response=None, url=None, body='', *args, **kwargs):
             self._matches.append(method_or_response)
             return
 
-        if method_or_response:
-            kwargs['method'] = method_or_response
+        self._matches.append(
+            Response(method=method_or_response, url=url, body=body, **kwargs))
+
+    def remove(self, method_or_response=None, url=None):
```
dcramer: i think it'd be better to require the instance on removal, and simply create an equality check on the BaseResponse class. Alternatively it needs to remove *any* matching instance.

responses.py
```diff
@@ -178,12 +183,14 @@ def get_headers(self):
     def get_response(self, request):
         raise NotImplementedError
 
-    def matches(self, request):
+    def matches(self, request, url_object_equality=False):
```
dcramer: if we need this here, per below, we should have it be a pure equality check (e.g. ``__eq__``), but if its going to rename on this method it should be something more generic like ``strict``.

responses.py
joscarsson: I implemented the equality check, and this now removes all matching instances (it still supports passing the instance, but it's pretty convenient given the use case in the PR description to be able to pass the method and url too).

joscarsson:
@dcramer: I've updated the PR according to your comments, I think it turned out pretty nice. The only thing that might be a bit weird is that `__eq__` only considers `method` and `url`, but the code in `remove` and `replace` is more readable now so I think it's worth it.

Please take a look :)

responses.py
dcramer: one last bit -- could we add a basic docstring to this to show syntax and usage

same for replace below

otherwise lgtm

joscarsson: @dcramer: Docstrings added, let me know what you think.

dcramer:
Thanks!

joscarsson:
Thanks for reviewing and merging @dcramer!

PR #123, opened 2016-08-23
Add section to README about using multiple responses

MichaelAquilina:
Describe how to use a secret feature in the responses library that I often (and otheres) find use for.

I created this since there doesnt seem to be any intention of merging #101 in, so it would be good to document how this is done without explicit support.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/responses/123)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -165,3 +165,25 @@ can be disabled by passing the ``assert_all_requests_are_fired`` value:
             rsps.add(responses.GET, 'http://twitter.com/api/1/foobar',
                      body='{}', status=200,
                      content_type='application/json')
+
+Multiple Responses
+------------------
+You can also use ``assert_all_requests_are_fired`` to add multiple responses for the same url:
+
+.. code-block:: python
+
+    import responses
+    import requests
+
+
+    def test_my_api():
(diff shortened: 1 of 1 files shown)


MichaelAquilina:
The failure on travis seems unrelated to my change? Its just that flake8 doesnt support python2.6 (it requires OrderedDict)

MichaelAquilina:
@dcramer (not sure if you're the right person to ping). Have you had a chance to look at this?

sgillies:
@MichaelAquilina I've proposed a fix for linting in #127.

dcramer:
@MichaelAquilina There's a lot of PRs/issues/etc on this repo, and to be quite honest the original API/intentions worked just fine. I'd be happy to add other contributors, but often people submit PRs which just greatly break behavior for others. Given that the project is fairly stable (albeit after I fixed the last set of half-broken changes), I haven't been overly eager to merge any other changes without full QA passes. Also given that my day job is becoming more and more demanding, I haven't had time to push forward many projects, including these one.

I will say, that I think there needs to be a 2.0 (or whatever version is next) that greatly rewrites some of the registration APIs to better take into account on-disk fixtures and simpler request matching for complex params as well as POST payloads.

PR #126, opened 2016-10-13
Adjust example for requests>2.10 compatibility

sgillies:
Test code of mine that executes `payload = json.loads(request.body)` in a responses callback, which was working with requests 2.10, has started failing after the release of requests 2.11.1 In requests 2.10 it seems that the `body` of a prepared request given to a responses callback could be a `str` but in 2.11.1 it is `bytes`. I've confirmed that `body.decode('utf-8')` is compatible with both requests 2.10.0 and 2.11.1.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -64,7 +64,7 @@ Request callback
     def test_calc_api():
 
         def request_callback(request):
-            payload = json.loads(request.body)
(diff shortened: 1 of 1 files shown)


sgillies:
If #127 is merged, or otherwise fixed, builds for this PR will pass.

PR #127, opened 2016-10-14
Skip `make lint` for Python 2.6

sgillies:
This PR skips linting for Python 2.6, which isn't supported by flake8 anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,5 +23,5 @@ install:
   - "pip install ${REQUESTS}"
   - make develop
 script:
-  - make lint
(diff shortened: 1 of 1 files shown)


MichaelAquilina:
:+1:

dcramer:
Ideally we'd lock the version, or just stop supporting Python 2.6, but I'm fine with this.

sgillies:
Thanks, @dcramer. I don't have any Python 2.6 in production and would prefer to get flake8 fixes, but can't speak for anyone else.

PR #129, opened 2016-11-08
Reraising exceptions when exiting context manager

dgouldin:
Currently when using responses as a context manager with `assert_all_requests_are_fired`, any exception raised inside the `with` block will be swallowed if not all requests have been fired. This can obscure issues in the code unrelated to its requests behavior.

This pull request bypasses responses' assertion that all requests were fired when exiting the context manager due to a raised exception.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -177,9 +177,11 @@ class RequestsMock(object):
         self.start()
         return self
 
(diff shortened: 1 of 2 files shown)


dcramer:
👍thanks

PR #131, opened 2016-12-14
Add remove_url_when_fired option

stragulus:
This option supports the model where one does want to make sure all mocked URL's are triggered at least once, while still allowing the same mocked URL from being called multiple times.

While it is possible to simply add the same mocked URL multiple times, it can be difficult in some uses cases to predict exactly how often the same URL gets called.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -119,10 +119,15 @@ class RequestsMock(object):
     POST = 'POST'
     PUT = 'PUT'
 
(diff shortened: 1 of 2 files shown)


dcramer:
I realize this is ancient, but I think what we want to do around this is take ``call_count`` from #145, and support something like "max number of responses" per match.

Issue #130, opened 2016-11-28
Reason is missing


whs:
The field [reason](https://docs.python.org/3/library/http.client.html#http.client.HTTPResponse.reason) is missing from the mocked request. It is usually set to HTTP status code's readable name (eg. "OK" "Not Found")

This make some libraries, like [requests_toolbelt dumping](https://toolbelt.readthedocs.io/en/latest/dumputils.html) unhappy.

whs:
See pull request #132

PR #132, opened 2016-12-15
Added reason to mocked HTTPResponse

whs:
See #130 

> The field reason is missing from the mocked request. It is usually set to HTTP status code's readable name (eg. "OK" "Not Found")
> 
> This make some libraries, like requests_toolbelt dumping unhappy.


dcramer:
Thanks!

PR #139, opened 2017-02-15
allow to return a response with no content-type header

anentropic:
Recently I had to write some tests against an API endpoint which does not return a Content-Type header in the response.  This is bad practice on the part of the API in question of course, but currently it is impossible to simulate this with `responses`.


dcramer:
Thanks!

PR #141, opened 2017-02-23
docs(README.rst): describe other add params

njamaleddine:
# What does this add?
Describes additional parameters for `responses.add()`

# Why?
I ran into an issue where I was trying to match a url with query params and no match was being found. Upon looking into `responses.py` I realized the `add` method required `match_querystring=True` to be set for it to match.


droope:
I was going to add a PR for this same issue. It can be frustrating when responses silently ignores the query string so it would be good to highlight this behaviour on the documentation.

dcramer:
I fixed this (without noticing this PR) as part of the 0.6.0 push. Thanks for opening this though!

PR #150, opened 2017-06-07
add pass_through option

itajaja:
This option allows to bypass the patch if there is no matching url. The option is useful in a couple of scenarios.

1) if a library is using responses, a double patching could happen, with unwanted consequences (see eg https://github.com/spulec/moto/issues/948)
2) as part of an integration testing, where there are actual endpoints that can be accessed

Let me know if this sounds reasonable or I am missing a glaring issue :)

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -10,6 +10,7 @@ import six
 from collections import namedtuple, Sequence, Sized
 from functools import update_wrapper
 from cookies import Cookies
+from requests.adapters import HTTPAdapter
 from requests.utils import cookiejar_from_dict
 from requests.exceptions import ConnectionError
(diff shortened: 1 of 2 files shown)


itajaja:
I just searched for something similar, and stumbled on this old pr: https://github.com/getsentry/responses/pull/39. It's somewhat similar, but this is slightly different in that it solves issues like the moto one pointed above. moto is a fairly used library, and I'd be beneficial to support this.

itajaja:
thanks!

kevgliss:
@itajaja did #174 solve your issue with moto? As far as I can tell, it seems that moto still taking over any subsequent mocks.

PR #154, opened 2017-07-25
Implement streaming files

EngineerCoding:
This also includes a more general approach of handling the body from the match. Test added and all tests passes.


dcramer:
Thanks!

PR #155, opened 2017-07-25
Various improvements and refactor

dcramer:
```python
responses.add(Response('GET', 'http://example.com'))
```


dcramer:
Putting this up for feedback at a first pass in cleaning up the way you register matches. This makes it a lot more flexible and gets us to the point where you could do more dynamic fixtures.

PR #157, opened 2017-07-27
Fix README.rst syntax

ownaginatious:
Looks like someone used markdown syntax in an RST document. This corrects that so that the README doesn't look messed up.


dcramer:
thanks!

PR #164, opened 2017-08-04
fix: allow returning arbitrary status codes

bartekwojcicki:
HTTPResponse.reason can be left blank. There is no reason to fail
for status codes like 418. It worked before 0.6.


dcramer:
Thanks!

PR #166, opened 2017-08-08
Fix match querystring

bartekwojcicki:


bartekwojcicki:
In 9ce871f I assume match_querystring means match if _all expected params  and no other ones were given_.

bartekwojcicki:
Feel free to ping me if there is anything I can do to help ship 0.6.2.

PR #168, opened 2017-08-08
Fix bug with reusing mocks out of order

singingwolfboy:
Looks like responses 0.6.2 has a bug that shows up when you reuse mocks in a different order from how they are defined. Here's a test case to reproduce the bug, and a fix to make the test case pass.


dcramer:
Somehow this slipped cleanup, but there was another fix for this shipped (and the tests also reflect/pass), so I'm just going to close this out.

Issue #170, opened 2017-08-08
Multiple response URLs not working correctly


ngaya-ll:
The fix for #161 breaks multi-URL functionality.

Test case:
```
def test_multiple_urls():
    @responses.activate
    def run():
        responses.add(responses.GET, 'http://example.com/one', body='one')
        responses.add(responses.GET, 'http://example.com/two', body='two')

        resp = requests.get('http://example.com/two')
        assert_response(resp, 'two')
        resp = requests.get('http://example.com/one')
        assert_response(resp, 'one')

    run()
    assert_reset()
```

dcramer:
gonna get the regression test in ahead of a fix

ngaya-ll:
I have a fix, will post in next 5 min.

dcramer:
:+1: thanks!

bartekwojcicki:
I'm sorry for introducing this one, thank you both for the fix.

fauskanger:
[ex]: https://github.com/getsentry/responses#multiple-responses

I've used the [Multiple responses example][ex], and it works:
    
    @responses.activate
    def test_multiple_to_same_url():
        responses.add(responses.GET, 'http://twitter.com/api/1/foobar', status=500)
        responses.add(responses.GET, 'http://twitter.com/api/1/foobar',
                      body='{}', status=200,
                      content_type='application/json')
    
        resp = requests.get('http://twitter.com/api/1/foobar')
        assert resp.status_code == 500
        resp = requests.get('http://twitter.com/api/1/foobar')
        assert resp.status_code == 200

But if I rearrange so that the a request has been made to _the same_ url, between adding the response it doesn't work.


    @responses.activate
    def test_multiple_to_same_url():
        responses.add(responses.GET, 'http://twitter.com/api/1/foobar', status=500)
    
        resp = requests.get('http://twitter.com/api/1/foobar')
        assert resp.status_code == 500
    
    
        responses.add(responses.GET, 'http://twitter.com/api/1/foobar',
                      body='{}', status=200,
                      content_type='application/json')
        resp = requests.get('http://twitter.com/api/1/foobar')
        assert resp.status_code == 200

The error output:

    ================================== FAILURES ===================================
    __________________________ test_multiple_to_same_url __________________________
    
        @responses.activate
        def test_multiple_to_same_url():
            responses.add(responses.GET, 'http://twitter.com/api/1/foobar', status=500)
    
            resp = requests.get('http://twitter.com/api/1/foobar')
[...]

PR #171, opened 2017-08-08
Issue #170: Fix bug with handling multiple matches

ngaya-ll:
Fixes a bug (#170) where the list of matches is incorrectly updated after a match, leading to issues with multiple responses.

I'm not entirely sure why the list of matches is being rotated in the first place, so this may create some other unwanted side effect.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -386,13 +386,13 @@ class RequestsMock(object):
         return get_wrapped(func, _wrapper_template, evaldict)
 
     def _find_match(self, request):
-        for match in self._matches:
+        for i, match in enumerate(self._matches):
             if match.matches(request):
                 break
         else:
             return None
-        # shift matches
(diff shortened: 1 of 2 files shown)


dcramer:
@ngaya-ll we want to be able to define multiple matches for the same path so that e.g. when the first is hit, it then goes to the second

dcramer:
that said, actually the behavior should be:

when match is hit, if there's another acceptable match, remove the item from the list and dont reinsert it

This means you can:

- define a behavior that happens initially
- leave the 'final response' state as whatever you want.

ngaya-ll:
Ah, ok.  I can update the PR to achieve that behavior.

dcramer:
yeah lets do that

would you mind also updating the test?

dcramer:
thanks!

PR #177, opened 2017-10-16
Add support for unicode in domain names and tlds

kathawala:
Earlier versions would have URLs like http://www.संजाल.भारत/hi/वेबसाइट-डिजाइन result in a connection error. This PR fixes that by encoding unicode in domain names and TLDs correctly when matching URLs.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -11,7 +11,6 @@ import six
 from collections import namedtuple, Sequence, Sized
 from functools import update_wrapper
 from cookies import Cookies
-from publicsuffixlist import PublicSuffixList
 from requests.adapters import HTTPAdapter
 from requests.exceptions import ConnectionError
 from requests.sessions import REDIRECT_STATI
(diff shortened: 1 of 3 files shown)


responses.py
```diff
@@ -49,6 +50,8 @@ def wrapper%(signature)s:
 
 logger = logging.getLogger('responses')
 
+psl = PublicSuffixList()
```
dcramer: unused

responses.py
```diff
@@ -201,6 +217,8 @@ def _url_matches(self, url, other, match_querystring=False):
                 url = _clean_unicode(url)
                 if not isinstance(other, six.text_type):
                     other = other.encode('ascii').decode('utf8')
+                print(url)
```
dcramer: stray prints

responses.py
kathawala: Ugh, sorry about these mistakes. I've fixed them up and I'll be more vigilant next time. Thanks

PR #178, opened 2017-10-17
Add unicode support for passthru urls

kathawala:
Unicode URLs were not supported in passthru mode, so this fixes that.


kathawala:
@dcramer Should I close this PR? I think it is still merge-able...

Issue #124, opened 2016-08-24
README should contain pip install info


augustjd:
Small point, but it would be nice to know that `pip install responses` does install the right library from the README. Put it right at the top of the README like other libraries.

rmad17:
I had added this long back in #187

markstory:
Closing as this has been fixed. Thanks for the nudge @rmad17

PR #179, opened 2017-10-18
Add pip info to Readme, fix some formatting

kathawala:
Resolves issue #124 which seemed pretty popular. Just adds "pip install" info and fixes some stray backticks.


kathawala:
This was fixed in #187

Issue #135, opened 2017-01-03
Test Max Retries


HangingClowns:
I made an HTTP adapter:

```
session = requests.Session()
session.verify = False

# Try max of 3 times before erroring out
retry_adapter = requests.adapters.HTTPAdapter(max_retries=3)
session.mount('https://', retry_adapter)
```

I wanted to make sure it will automatically retry, but it seems it's not getting triggered right:

```
exception = ConnectionError('Bad connection')
with responses.RequestsMock(assert_all_requests_are_fired=False) as rsp:
   rsp.add(responses.POST, url, body=exception)
   rsp.add(responses.POST, url, body=exception)
   rsp.add(responses.POST, url, body='{}')
   session.post(url=url, json={})
```

It just keeps raising the exception.

tuukkamustonen:
Looks like `responses` mocks the `adapter.send()` call in `Session.send()` and re-routes it into `RequestsMock`, that just doesn't support retrying...

lsh-0:
just encountered this myself

cjw296:
me too, doesn't look like anyone from Sentry is monitoring this issue tracker though...

kathawala:
I have a `max-retries` branch on my fork (https://github.com/kathawala/responses/tree/max-retries) which allows this to run without raising an exception:

```python
import responses
import requests

exception = ConnectionError('Bad connection')
url = 'https://www.example.com'
with responses.RequestsMock(assert_all_requests_are_fired=False, max_retries=3) as rsp:
   rsp.add(responses.POST, url, body=exception)
   rsp.add(responses.POST, url, body=exception)
   rsp.add(responses.POST, url, body=exception)
   rsp.add(responses.POST, url, body='{}')
   requests.post(url=url, json={})
```

But what's the use case for this? Just want to know before I submit a PR. It's not clear to me how this could be helpful.

cjw296:
@kathawala - I configured a `max_retries` as `Retry()` instance and I wanted to check it would work as I expected. Would your branch help with that?

kathawala:
@cjw296 Hmm, I'm not exactly sure what you mean. What do you mean you "configured a `max_retries` as `Retry()` instance"? Code examples would help me understand :)

cjw296:
here's the code I'd like to test:
```python
from requests.adapters import HTTPAdapter
from requests.packages.urllib3 import Retry
...
        retry = Retry(
            total=5,
            status=3,
            status_forcelist=[404],
[...]

PR #180, opened 2017-10-21
Add max_retries option to RequestsMock

kathawala:
Issue #135 shows a number of people who want to use `responses` to mock requests to a website and retries of that request to the same website.

Not sure why people want to do this, but if it's something desirable, this should allow for that. Usage would look like this (script will run without raising a `ConnectionError`.

```python
import responses
import requests

exception = ConnectionError('Bad connection')
url = 'https://www.example.com'
with responses.RequestsMock(assert_all_requests_are_fired=False, max_retries=3) as rsp:
   rsp.add(responses.POST, url, body=exception)
   rsp.add(responses.POST, url, body=exception)
   rsp.add(responses.POST, url, body=exception)
   rsp.add(responses.POST, url, body='{}')
   requests.post(url=url, json={})
```

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -560,6 +560,7 @@ class RequestsMock(object):
             import mock
 
         def unbound_on_send(adapter, request, *a, **kwargs):
+            retries = adapter.max_retries if isinstance(adapter.max_retries, int) else adapter.max_retries.total
             return self._on_request(adapter, request, adapter.max_retries.total, *a, **kwargs)
 
         self._patcher = mock.patch('requests.adapters.HTTPAdapter.send',
--- a/responses.py
+++ b/responses.py
@@ -341,14 +341,12 @@ class RequestsMock(object):
     def __init__(self,
                  assert_all_requests_are_fired=True,
                  response_callback=None,
-                 passthru_prefixes=(),
-                 max_retries=0):
+                 passthru_prefixes=()):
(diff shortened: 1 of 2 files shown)


dcramer:
I would also like to understand the use case here before adding the feature. Even if we were to add it, I think it should simply use request's internal max retries (per request?). That said, I feel like "does the HTTP adapter retry" is not really in scope for this project.

kathawala:
> Even if we were to add it, I think it should simply use request's internal max retries (per request?).

Yes, that's more reasonable. I changed it to do that.

> That said, I feel like "does the HTTP adapter retry" is not really in scope for this project.

@dcramer 
I really only opened this in case it was something you all at Sentry saw as a desirable feature. If that's not the case, I can close it, and issue #135 can be closed as well.

conradogarciaberrotaran:
Is this PR dead for some reason? max_retries support would be a good feature here..

markstory:
@conradogarciaberrotaran It hasn't made any progress because the use cases was a bit murky to start with and (still are). Additionally, the changes can no longer be merged.

conradogarciaberrotaran:
Thanks for the response Mark. For anyone googling, I went with HTTPretty to test retries mechanics: https://httpretty.readthedocs.io/en/latest/

PR #181, opened 2017-10-27
Fix README examples with import of requests library

feliperuhland:
Added import of requests library.


dcramer:
Thanks, and sorry for the delay!

Issue #188, opened 2018-02-18
Unable to intercept vendored requests


wimglenn:
The path to the requests adapter is [hardcoded](https://github.com/getsentry/responses/blob/11e870728a30ca31885373bcbacb11fc32783a4d/responses.py#L564):

    self._patcher = mock.patch('requests.adapters.HTTPAdapter.send',
                               unbound_on_send)

However many projects vendor `requests`, and so their network activity just flies straight past the `RequestsMock` unnoticed.  To give an example, pip uses a vendored requests session and responses can not mock it out unless source code is modified with the correct mock location:

    self._patcher = mock.patch('pip._vendor.requests.adapters.HTTPAdapter.send',
                               unbound_on_send)

Could you add the requests path as configurable in the public API somehow?  Thanks!

QasimK:
This is also an issue if you use your own `HTTPAdapter`:(

thiagokokada:
@wimglenn I have a use case similar to you and I am trying to understand how can I use a vendored version of requests in a project. Can you give some examples?

PR #189, opened 2018-02-19
add support for custom patch target

wimglenn:
resolves https://github.com/getsentry/responses/issues/188

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -35,6 +35,11 @@ if six.PY2:
 else:
     from io import BytesIO as BufferIO
 
+try:
+    from unittest import mock as std_mock
+except ImportError:
+    import mock as std_mock
+
 UNSET = object()
 
 Call = namedtuple('Call', ['request', 'response'])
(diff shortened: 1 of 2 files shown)


wimglenn:
So it seems `assert_called_once` is [new in Python 3.6](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once) and is the cause of test failure for 3.5.  Fixed that.  The other failures were not related to my changes, but I addressed them in a separate PR.  

https://github.com/getsentry/responses/pull/190

responses.py
```diff
@@ -341,12 +346,14 @@ class RequestsMock(object):
     def __init__(self,
                  assert_all_requests_are_fired=True,
                  response_callback=None,
-                 passthru_prefixes=()):
+                 passthru_prefixes=(),
+                 target='requests.adapters.HTTPAdapter.send'):
```
dcramer: might make sense to just specify the base library for the target? (e.g. requests vs urllib3.vendor.requests)

dcramer:
@wimglenn you should generally never use 'assert_called_once', but rather always use 'assert_called_once_with' because the arguments being passed are important

responses.py
wimglenn: Hmm, I considered that, but couldn't think of any advantage over giving the full control to the client.  For example, somebody may have forked requests and/or subclassed HTTPAdapter, for whatever reason, and there isn't reason to prevent those users from customizing their target correctly.

Would you like me to change it?

wimglenn:
@dcramer   Regarding the preference of `assert_called_once_with` over `assert_called_once`, yes, generally I would agree.  But in this case, I couldn't see how to actually use it here, since the second argument required is a function hidden inside the closure of another function.  I don't need to assert on the second argument.  If you know a better way than the workaround I've put, then I'd be interested to see it!

QasimK:
I also need the capability to customise the patch target because I use my own `HTTPAdapter` instance (not sub-classed). Is there anything that I can do to move this pull request along and get another release of responses out?

dcramer:
test failures are unrelated to this change so merging in

PR #190, opened 2018-02-19
don't access private attributes of re

wimglenn:
Accessing the `re._pattern_type` seems to be failing the tests in nightly and 3.7-dev.


markstory:
Thanks for taking the time to put this together, but I'm going to close it as we have had another approach merged and tests are currently passing against Python 3.7

PR #193, opened 2018-03-09
Add support for Python 3.7

purple4reina:
Python 3.7 renamed [`re._pattern_type`](https://github.com/python/cpython/blob/v3.6.4/Lib/re.py#L283) to [`re.Pattern`](https://github.com/python/cpython/blob/v3.7.0b2/Lib/re.py#L268).

Also add testing envs for `py36` and `py37`.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -35,6 +35,12 @@ if six.PY2:
 else:
     from io import BytesIO as BufferIO
 
+if hasattr(re, '_pattern_type'):
+    _re_pattern_type = re._pattern_type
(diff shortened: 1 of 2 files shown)


wimglenn:
@purple4reina   this is also addressed in earlier https://github.com/getsentry/responses/pull/190

purple4reina:
@wimglenn Ah yes, your solution is much superior to mine! I'll go ahead and close this PR. Thanks so much!

PR #199, opened 2018-04-04
build: Overhaul CI and autoformatting

dcramer:
- Switch to Black
- Add VS Code configuration
- Apply Black to all files

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,8 +10,5 @@ repos:
     hooks:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
--   repo: git://github.com/pre-commit/pre-commit-hooks
-    rev: v0.9.1
-    hooks:
     -   id: debug-statements
     -   id: flake8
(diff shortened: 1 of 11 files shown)


.pre-commit-config.yaml
```diff
@@ -1,14 +1,17 @@
--   repo: git://github.com/pre-commit/mirrors-yapf
-    sha: v0.16.2
+repos:
+-   repo: https://github.com/ambv/black
+    rev: 51b3b2624d4b1e4608eb9addd7d292cc5174a047
     hooks:
-    -   id: yapf
+    - id: black
+      args: [--line-length=88, --safe]
+      python-version: python3.6
```
asottile: note you'll want `python_version` -- I updated the docs in https://github.com/ambv/black/pull/106 but I was not quick enough!

.pre-commit-config.yaml
```diff
@@ -1,14 +1,17 @@
...
 -   repo: git://github.com/pre-commit/pre-commit-hooks
-    sha: v0.9.1
+    rev: v0.9.1
     hooks:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 -   repo: git://github.com/pre-commit/pre-commit-hooks
-    sha: v0.9.1
+    rev: v0.9.1
     hooks:
```
asottile: you can actually remove these three lines and it'll work the same.  pre-commit lets you duplicately list repositories 🤷‍♂️

PR #207, opened 2018-05-29
Fix pass-thru requests losing additional settings

DHager:
This one is pretty straightforward, some request-settings are not getting passed along for stuff like proxies and certificates.

In my particular use-case the fix is essential because a SOCKS proxy is the only way to access the external resource. (Without `socks5h://` I can't even resolve the internal hostname.)


dcramer:
Not really sure what's going on with Travis, but looks green so gonna ignore check.

Thanks!

PR #209, opened 2018-06-12
Alow extra keyword arguments to be passed to the callback function

jobec:
Sometimes you need to register the same callback function multiple time, but with slightly different behavior (ex. a different username in the response). Being able to pass arguments to the callback function makes for less code duplication.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -162,6 +162,27 @@ a tuple of (``status``, ``headers``, ``body``).
             '728d329e-0e86-11e4-a748-0c84dc037c13'
         )
 
+If you want to pass extra keyword arguments to the callback function, for example when reusing
+a callback function to give a slightly different result, you can use ``functools.partial``:
+
+.. code-block:: python
+
+    from functools import partial
+
+    ...
+
+        def request_callback(request, id=None):
+            payload = json.loads(request.body)
+            resp_body = {'value': sum(payload['numbers'])}
+            headers = {'request-id': id}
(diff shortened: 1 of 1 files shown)


dcramer:
I'm ok with this change, but one thought:

Why not just use functools.partial?

I didn't put too much effort into the above, but I feel like it might work without adding new apis:

``responses.add_callback(responses.GET, url, partial(callback, extra={...}))``

jobec:
is `because I didn't know about partial()` a good answer? 😄 
I tested it and it indeed works.

Might be nice to add to the docs then.

jobec:
I've reverted my changes and instead added the use of `partial()` to the readme

README.rst
```diff
@@ -162,6 +162,27 @@ a tuple of (``status``, ``headers``, ``body``).
...
+
+        def request_callback(request, id=None):
+            payload = json.loads(request.body)
+            resp_body = {'value': sum(payload['numbers'])}
+            headers = {'request-id': id}
+            return (200, headers, json.dumps(resp_body))
+
+        responses.add_callback(
+            responses.POST, 'http://calc.com/sum',
+            callback=partial(request_callback, id='728d329e-0e86-11e4-a748-0c84dc037c13',
```
dcramer: actually mind fixing up this syntax real quick?

jobec:
Done

dcramer:
Thanks!

PR #210, opened 2018-06-28
Install test extras in tox

rouge8:
Without this, running `tox` fails because none of the test dependencies are installed.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -3,9 +3,6 @@
 envlist = {py26,py27,py32,py33,py34,py35}
 
 [testenv]
-deps =
-    pytest
-    pytest-cov
-    pytest-flakes
+extras = tests
 commands =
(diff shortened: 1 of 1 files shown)


rouge8:
it tells tox to install your `tests` extras: https://github.com/getsentry/responses/blob/master/setup.py#L26-L37

tox.ini
```diff
@@ -3,6 +3,7 @@
 envlist = {py26,py27,py32,py33,py34,py35}
 
 [testenv]
+extras = tests
 deps =
```
adamchainz: `deps` should be dropped since they all come from `extras`!

rouge8: good call, fixed ✅

markstory:
Thank you 👍

Issue #211, opened 2018-07-27
Response with query string, but match_querystring=False should raise a warning


rbu:
## Problem
I've debugged an unrelated part of the an application for hours before noticing that mocked responses were matched unexpectedly, due to me not seeing the default `match_querystring=False`. A naive test case might do something like this:

```py
$ python
Python 3.6.6 (default, Jul 19 2018, 14:25:17) 
[GCC 8.1.1 20180712 (Red Hat 8.1.1-5)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import responses
>>> import requests
>>> 
>>> responses.start()
>>> 
>>> responses.add(responses.GET, "https://example.com/api?endpoint=1",
...     json={'endpoint': 1}
... )
>>> responses.add(responses.GET, "https://example.com/api?endpoint=2",
...     json={'endpoint': 2}
... )
>>> 
>>> requests.get('https://example.com/api?endpoint=2').content
b'{"endpoint": 1}'
>>> requests.get('https://example.com/api?endpoint=1').content
b'{"endpoint": 2}'
>>> 
```

## Proposal
While I'd prefer the default for `match_querystring` to be changed, I'd propose responses throws a `RuntimeWarning` when a response is added that contains a query string without setting `match_querystring` explicitly.

rbu:
Another example of someone stumbling over this is #203

dcramer:
Not opposed to that. I'd also be fine just changing match_querystring=True (implicit only) when you pass a URL with a querystring.

rbu:
@dcramer I like the idea of a giving `match_querystring` a new default value that depends on the url. Is this something you'd be working on? If not, I'd look into submitting a PR for this next week.

dcramer:
I probably won’t have time, but if you throw up a PR I’m happy to review/merge

rbu:
@dcramer I have added a pull request for the automatic behavior you proposed in #213.
I'd very much appreciate your review.

PR #213, opened 2018-08-03
Change match_querystring default to consider URL

rbu:
Previously, adding a response with a query string without also setting
match_querystring=True would provide confusing results, see #211.

This commit changes the argument's default value to consider the URL and
automatically enable match_querystring if a query string is detected in
the response.


dcramer:
Thanks!

PR #215, opened 2018-09-08
Import collections ABCs from collections.abc if possible

jwodder:
"`from collections import Sequence, Sized`" produces a `DeprecationWarning` under Python 3.7 with the message:

> Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working

This patch eliminates this warning by importing from `collections.abc` instead, falling back to `collections` only if that fails (i.e., if running under Python < 3.3).

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -7,13 +7,18 @@ import logging
 import re
 import six
 
-from collections import namedtuple, Sequence, Sized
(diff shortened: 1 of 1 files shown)


markstory:
Thank you 👍

Issue #186, opened 2017-12-07
cookies pkg has a deprecation warning


bcail:
If you run a test script that just does "import responses" with the -Werror option, a DeprecationWarning exception is raised from the cookies package.

I'm not sure the cookies package is really maintained anymore (last commit in Sep 2014): https://gitlab.com/sashahart/cookies/commits/master. Maybe the biscuits pkg would be helpful? https://gitlab.com/sashahart/cookies/issues/4#note_50409121.

dcramer:
If someone wants to try to swap it out thats probably fine, but to me its a not a huge concern. You can always filter out warnings from the responses package if need be.

jsheedy:
I created pull request https://github.com/getsentry/responses/pull/218 which uses biscuits on Python 3 and continues to use cookies on Python 2.

PR #218, opened 2018-09-27
use biscuits on Python 3, cookies on Python 2

jsheedy:
This pull request addresses issue #186 by using the [biscuits](https://github.com/pyrates/biscuits) package instead of [cookies](https://gitlab.com/sashahart/cookies/) when on Python >= 3.4. Biscuits does not support Python < 3.4 so the original behavior is maintained there.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -107,12 +107,14 @@ def _is_redirect(response):
 
 
 def _cookies_from_headers(headers):
-    if sys.version_info[:2] < (3,4):
+    if sys.version_info[:2] < (3, 4):
         from cookies import Cookies
+
         resp_cookies = Cookies.from_request(headers["set-cookie"])
         cookies_dict = {v.name: v.value for _, v in resp_cookies.items()}
     else:
         import biscuits
+
         cookies_dict = biscuits.parse(headers["set-cookie"])
     return cookiejar_from_dict(cookies_dict)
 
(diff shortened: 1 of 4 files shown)


responses.py
```diff
@@ -25,6 +25,11 @@
     from urllib import quote
 else:
     from urllib.parse import urlparse, parse_qsl, urlsplit, urlunsplit, quote
+    
+if sys.version_info[:2] < (3,4):
+    from cookies import Cookies
+else:
+    import biscuits
```
markstory: Couldn't this also be 

```python
try:
    import biscuits
    use_biscuits = True
except ImportError:
    use_biscuits = False
    from cookies import Cookies
```

That would let you reduce the number of places you check the python runtime version, as you could use `if use_biscuits` instead.

An alternative approach would be for responses to provide an abstraction around the two libraries and handle the differences there.

jsheedy: Yes, I like the abstraction approach. I'll make an update.

jsheedy:
I moved the cookie handling to _cookies_from_headers() as reviewed. The imports are inside the function instead of at the top of the file, but that departure from standard seemed worth the tradeoff of abstracting the differences inside this function. The Travis builds are failing for Python 3.3 which has reached it's end-of-life date. The Python 3.6 Travis run is failing, though it shows the tests are passing as was my experience in testing. Does Travis run successfully on master?

PR #221, opened 2018-10-20
Update changelog for 0.10.0 and 0.10.1

adamchainz:
This was missed, I couldn't tell what was changing as I was upgrading - some of my tests broke - so I had to read the git log to figure this out. THis will help future travellers.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,15 @@
+0.10.1
+------
+
+- Packaging fix to distribute wheel (#219)
+
+0.10.0
+------
+
+- Fix passing through extra settings (#207)
+- Fix collections.abc warning on Python 3.7 (#215)
+- Use 'biscuits' library instead of 'cookies' on Python 3.4+ (#218)
+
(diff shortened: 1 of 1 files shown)


markstory:
What broke in your tests?

adamchainz:
`biscuits` wasn't installing because I compile my test requirements on Python 2 for both Python 2 and 3 - a bit of a hack I know, but in practicality this is the first time it has been a problem, normally there are just Python 2-only requirements (across like 20 projects). https://github.com/adamchainz/ec2-metadata/pull/96/files .

Not `responses`' fault, but on the other hand I did switch to `requests-mock` after making these PR's to `responses` and seeing how it works, and re-researching `requests-mock`. 🙈

markstory:
Thanks for the additional context @adamchainz I'll look for another way to parse set-cookie headers that doesn't introduce py3 only dependencies.

adamchainz:
There's `http.cookies` in the standard library

PR #224, opened 2018-10-23
Don't use a dependency to parse cookies in py3

markstory:
Instead of adding a dependency to parse cookies in python3 we can use stdlib features.


adamchainz:
👍

PR #229, opened 2018-11-08
Update inspection to be compatible with newer Python versions

### LINKED ISSUE #134: PY3.5: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() instead
While running responses in Python 3.5 environment, I can see now a DeprecationWarning.

```
DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() instead
```

I know this is not critical at the moment, but may be a major issue in the future when Python 3.6 will be more popular (stable 3.6.0 release is available since 23.12.2016).

Timvde:
`getargspec` and `formatargspec` are deprecated since Python 3.5 and
produce deprecation warnings. Since Python 3.3, a new inspection API has
been available. We use it starting from Python 3.4, so we can make use
of the existing six.PY34 constant.

Fixes #134.

I welcome any hints to make this code cleaner, I don't have a lot of experience with inspection.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -129,18 +129,27 @@ def _cookies_from_headers(headers):
 def get_wrapped(func, wrapper_template, evaldict):
     # Preserve the argspec for the wrapped function so that testing
     # tools such as pytest can continue to use their fixture injection.
-    if six.PY2:
-        args, a, kw, defaults = inspect.getargspec(func)
-    else:
-        args, a, kw, defaults, kwonlyargs, kwonlydefaults, annotations = inspect.getfullargspec(
-            func
-        )
-
-    signature = inspect.formatargspec(args, a, kw, defaults)
     is_bound_method = hasattr(func, "__self__")
-    if is_bound_method:
-        args = args[1:]  # Omit 'self'
-    callargs = inspect.formatargspec(args, a, kw, None)
+    if six.PY34:
+        full_signature = inspect.signature(func)
+        parameters = full_signature.parameters.values()
+        signature = str(full_signature)
+        if is_bound_method:
+            signature = signature[:1] + 'self, ' + signature[1:]
+        callargs = str(full_signature.replace(
+            parameters=(inspect.Parameter(p.name, p.kind) for p in parameters))
+        )
+    else:
+        if six.PY2:
+            args, a, kw, defaults = inspect.getargspec(func)
+        else:
+            args, a, kw, defaults, kwonlyargs, kwonlydefaults, annotations = inspect.getfullargspec(
+                func
+            )
+        signature = inspect.formatargspec(args, a, kw, defaults)
+        if is_bound_method:
+            args = args[1:]  # Omit 'self'
+        callargs = inspect.formatargspec(args, a, kw, None)
 
     ctx = {"signature": signature, "funcargs": callargs}
     six.exec_(wrapper_template % ctx, evaldict)


markstory:
Thank you for taking the time to do this. There was another open pull request for this (#222) that has been merged.

Timvde:
I am not convinced that the merged code is correct. In particular this line: https://github.com/getsentry/responses/pull/222/files#diff-9206e1daa98d75bd8b5c902a12aa5ea5R143

[PEP-362](https://www.python.org/dev/peps/pep-0362/#implementation) says:

> If the object is a bound method, construct and return a new Signature object, with its first parameter (usually self or cls) removed. (classmethod and staticmethod are supported too. Since both are descriptors, the former returns a bound method, and the latter returns its wrapped function.)

This means that instead of removing the first argument in the function call arguments, `self` should be *added* to the wrapper arguments.

Also, `inspect.signature` is [new since Python 3.3](https://docs.python.org/3/library/inspect.html#introspecting-callables-with-the-signature-object), so I expect the code to be broken for Python 3.0, 3.1 and 3.2 (which is not tested by the CI though, so I'm not sure if those versions are still supported).

markstory:
>  I expect the code to be broken for Python 3.0, 3.1 and 3.2 (which is not tested by the CI though, so I'm not sure if those versions are still supported).

Python 3.0, 3.1 and 3.2 are not supported.

There are tests covering bound methods ensuring that self is included in the argspec [here](https://github.com/getsentry/responses/pull/222/files#diff-a95e1a562f3879ddb72fe81003f48c52R535). Are there scenarios we're missing?

Timvde:
I debugged the test case, and it seems that `is_bound_method` is `False`. I'm not sure what the consequences of that are (I'm not that familiar with the code). You can see in the code coverage tool that the block isn't hit: https://codecov.io/gh/getsentry/responses/pull/222/diff?src=pr&el=tree#D2-143

markstory:
I would have expected `is_bound_method` to be True as the tests use instance methods (which do have `__self__`). I'll take another look.

Timvde:
I added another commit on my branch: https://github.com/Timvde/responses/commit/441f2a8a866c15a2edf1c9c4b6bde7cfaf604105

There is another issue that the return type annotation doesn't get removed. I also removed Python 3.1-3.3 compatibility. If it's unsupported, we don't need to clutter the code with it.

If you agree with my assessment, let me know, and I'll rebase and create a new PR.

Did you already find out why `is_bound_method` is `False`?

Timvde:
My code now fully passes our project's test suite; current master doesn't. Error:

```
.venv/lib/python3.7/site-packages/responses.py:158: in get_wrapped                                                                                                                                                   
    evaldict,                                                                                                                                                                                                                                                            
E     File "<string>", line 3                                                                                                                                                                                                                                                                                                
E       return func() -> None                                                                                                                                                                                                                                                                                                
E                      ^                                                                                                                                                                                                                                                                                                     
E   SyntaxError: invalid syntax
```

markstory:
@Timvde I've opened #231 to fix return type hints as well.

Timvde:
You also need to remove the types from the `wrapper_args`. I got import errors if I left them there.

Apart from that, this line still cannot be correct: https://github.com/getsentry/responses/pull/231/files#diff-9206e1daa98d75bd8b5c902a12aa5ea5R145. Have you looked into that?

markstory:
> You also need to remove the types from the wrapper_args. I got import errors if I left them there.

You're right, I'll get my pull request updated.

> Have you looked into that?

Yes, I can't find any cases where that value is true in py3, so I'm going to remove it from the py3 code path. I've updated #231 with additional changes and companion tests.

Timvde:
Hmm, it seems that you are right. If I put a breakpoint in the code there, it never hits. Now I wonder how this will actually work on bound methods. Wouldn't you need `self` in the signature?

Apart from the return type, you forgot to also remove type hints in the argument here: https://github.com/getsentry/responses/pull/231/files#diff-9206e1daa98d75bd8b5c902a12aa5ea5R142. Our tests still fail with a `NameError` because the module name is not imported. It does seem that mypy can still do type checking, even with the type signature removed, so this should have no negative effects.

Apart from these remarks, I think your patch is good to go.

markstory:
The return annotation should be removed in [line 142](https://github.com/getsentry/responses/pull/231/files#diff-9206e1daa98d75bd8b5c902a12aa5ea5R142). Do you have an example function that does not work with the current patch? I'd like to include it in our test suite.

Timvde:
Type hints in the arguments should also be stripped. So the function

```pyton
def myfunction(foo: Foo): pass
```

with Foo some custom type, doesn't work. Your replace should look like this: https://github.com/Timvde/responses/commit/b10ccea70f0809bead6c09df6cdca38a93c50e5c#diff-9206e1daa98d75bd8b5c902a12aa5ea5R143

markstory:
@Timvde Thanks. Using a non-built-in type for the parameter annotation was the bit I was missing.

Timvde:
Passes our test suite without any errors or warnings now. We don't use any responses decorator on bound methods, so although I'm still not convinced that that part of the code is correct now, it would not impact us and the code is good to go for me. Could you tag a new version after releasing please?

PR #235, opened 2018-12-05
Fix callback exception mocking.

### LINKED ISSUE #234: callback results are not unpacked prior to checking for exception bodies
This means the way to mock an error for a callback differs from the way to mock andd error for standard use. Presumably this is not desired - it is not documented at any rate.

jsirois:
It now works like regular response mocking where a body that
is an Exception subclass triggers that exception to be raised.
Support for the existing undocumented callback-only behavior
is retained but a deprecation warning issued. Tests are added
for both of these cases.

Fixes #234

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -6,6 +6,7 @@ import json as json_module
 import logging
 import re
 import six
+import warnings
 
 from collections import namedtuple
 from functools import update_wrapper
@@ -387,9 +388,19 @@ class CallbackResponse(BaseResponse):
 
         result = self.callback(request)
         if isinstance(result, Exception):
+            warnings.warn(
+                "Callbacks should always return a (status, headers, body) "
+                "tuple. To mock an exception, provide that exception as "
+                "the body in the tuple. Support for exception responses "
+                "from callbacks will be removed in version 1.0.0.",
+                DeprecationWarning,
+            )
             raise result
 
         status, r_headers, body = result
+        if isinstance(body, Exception):
+            raise body
+
         body = _handle_body(body)
         headers.update(r_headers)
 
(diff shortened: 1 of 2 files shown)


responses.py
```diff
@@ -387,9 +388,19 @@ def get_response(self, request):
 
         result = self.callback(request)
         if isinstance(result, Exception):
+            warnings.warn(
```
jsirois: I presumed the right answer here was aligning callback exception handling to regular exception handling. It may be that you prefer the opposite direction (I'm guessing unlikely) or that you intend this difference and would rather just have a doc update PR (more likely). Just let me know if this approach is off target or else too heavyweight - perhaps due to 0.X.Y demver status you're ready to just break API users and kill this style outright.

dcramer: We were emulating another API where you simply return exceptions as the result, which is ultimately what this was doing. That said, we should either raise, or return, and not both. I do think it's kind of silly to return an exception *and* status/headers, since those don't affect the exception. To me it just feels like extra code that doesn't provide value (as it doesn't affect behavior). Open to having my mind changed though. (I'm also open to handling the case where you return an Exception in the body, and not deprecating the other mechanism)

jsirois: I think I agree with returning either Exception or (status, headers, body).

Before I take a second stab, please confirm this course of action:
1. Update the "normal" API to accept either Exception or (status, headers, body) and deprecate body as exception in that API.
2. Revert this change to the callback API since it already implements the preferred API.
3. Update README to reflect the new preffered API and mention the deprecated body as Exception support.

jsirois: Hrm, let me look at the APIs a bit more. Bringing the "normal" API in-line with the undocumented callback API looks like it requires care.

jsirois: I'll do that. I'd personally be sensitive to maintaining two api's vs deprecating the documented / wonky one - but since you're ok with it, I'm on it. Thanks for the feedback.

jsirois: Fixed.

dcramer:
Thanks!

PR #239, opened 2018-12-14
Match supported version in Travis, tox, and trove classifiers

jdufresne:
In Travis, use 'dist: xenial' to allow testing the Python 3.7 release.
Python 3.7 was released on June 27, 2018.

https://docs.python.org/3/whatsnew/3.7.html

In tox.ini, test the same test matrix as Travis. This allows
contributors to easily test all supported environments locally.

In setup.py, document these same versions as support Python versions.
This allows new users to quickly see, at a glance, if this project is
suitable as a dependency. These values are also used by scripts.

Include python_requires in setup.py to help pip decide what version of
the library to install. As the support Python versions change, pip can
use this option.


setup.py
```diff
@@ -85,8 +86,15 @@ def run_tests(self):
         "Intended Audience :: Developers",
         "Intended Audience :: System Administrators",
         "Operating System :: OS Independent",
+        "Programming Language :: Python",
         "Programming Language :: Python :: 2",
+        "Programming Language :: Python :: 2.7",
         "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3",
```
jdufresne: Ooops. I duplicated "3" here, will fix.

jdufresne:
Fixed duplicate classifier.

PR #249, opened 2019-01-08
removed extra comma in docs

jameslamb:
@getsentry I've really been digging `responses`! I've found it easier to work with than `unittest.mock`. Thanks for providing it!

Please consider this small PR to remove a trailing comma I noticed while going through the README.


markstory:
Thank you 👍

PR #250, opened 2019-01-16
Add documentation on regex-matching

scubbo:
Earlier documentation does state that a compiled regex can
be passed to the `url` argument, but it's easy to miss.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -196,7 +196,7 @@ You can also pass a compiled regex to `add_callback` to match multiple urls:
 
         responses.add_callback(
             responses.POST,
-            re.compile('http://calc.com/[sum|prod|pow|unsupported]'),
+            re.compile('http://calc.com/(sum|prod|pow|unsupported)'),
             callback=request_callback,
             content_type='application/json',
         )
--- a/README.rst
+++ b/README.rst
@@ -164,6 +164,53 @@ a tuple of (``status``, ``headers``, ``body``).
             '728d329e-0e86-11e4-a748-0c84dc037c13'
         )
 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -164,6 +164,92 @@ a tuple of (``status``, ``headers``, ``body``).
...
+
+        try:
+          resp = requests.post(
+                'http://calc.com/foobar',
+                json.dumps({'numbers': [2, 3, 4]}),
+                headers={'content-type': 'application/json'},
+          )
+          assert False, "Expecting an exception here"
+        except requests.exceptions.ConnectionError:
+          print('Test of unhandled path passed')
```
markstory: Could we include only one sample usage instead of all the permutations? This snippet is a bit long right now.

scubbo: Sure thing! Will amend.

README.rst
```diff
@@ -164,6 +164,53 @@ a tuple of (``status``, ``headers``, ``body``).
...
+
+            operator = operators[operator_name]
+
+            resp_body = {'value': reduce(operator, payload['numbers'])}
+            headers = {'request-id': '728d329e-0e86-11e4-a748-0c84dc037c13'}
+            return (200, headers, json.dumps(resp_body))
+
+        responses.add_callback(
+            responses.POST,
+            re.compile('http://calc.com/[sum|prod|pow|unsupported]'),
```
markstory:
```suggestion
            re.compile('http://calc.com/(sum|prod|pow|unsupported)'),
```

I think you want an alternating group not a character class. Right now your regex captures `http://calc.com/mus`

Issue #252, opened 2019-01-22
View/modify registered responses?


ptink:
I get the sense that I'm not using the library correctly but I have a list of several "default" responses that I want to be registered for most test cases. I'm using the `ResponsesMixin` from [here](https://github.com/getsentry/responses/issues/31) but this comes with its own set of issues for example if I want to override just one of the registered responses for a test there doesn't seem to be a good way to do that- also I could not find a way to inspect the list of all currently registered responses inside a test.

Is this sort of behavior outside the scope of this project or is it possible and the documentation could be expanded to include this?

markstory:
You can modify the list of mocked responses in a few ways. You can use `replace()` to overwrite an existing mock, and `remove()` to remove a mock based on method and URL. Lastly you can use `clear()` to reset all mocks. There aren't any public APIs for looking at the currently connected mocks.

PR #253, opened 2019-01-23
Added matches property to RequestMock. Updated documentation for modifying responses

ptink:
Referencing https://github.com/getsentry/responses/issues/252#event-2089058954 this PR adds a property that can be used to check the `_matches` property on the `RequestsMock`. Also updated the documentation to add some detail around replacing/removing registered responses

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -553,6 +553,10 @@ class RequestsMock(object):
     def calls(self):
         return self._calls
 
+    @property
+    def matches(self):
+        return self._matches
+
     def __enter__(self):
         self.start()
         return self
--- a/responses.py
+++ b/responses.py
@@ -553,10 +553,6 @@ class RequestsMock(object):
     def calls(self):
         return self._calls
 
-    @property
-    def matches(self):
-        return self._matches
-
     def __enter__(self):
         self.start()
         return self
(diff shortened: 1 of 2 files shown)


responses.py
```diff
@@ -553,6 +553,10 @@ def add_callback(
     def calls(self):
         return self._calls
 
+    @property
+    def matches(self):
```
markstory: Why not a standard function? That would be more consistent with the other parts of responses and fit with the module auto-exporter.

ptink: I was matching the implementation of `calls` above, as they seemed to be similar grouping of functionality. On review I worry that `matches` is too confusing a name because ``BaseResonse`` already has a `matches()` function. Do you have a preferred name or method signature?

markstory: Do we need a public method? Are you frequently dumping the mocks you created?

ptink: I found it useful for debugging. Especially when initially getting used to the library I found I was getting issues with trying to call a response that I thought I had added but had no way to be sure. Being able to check the `_matches` greatly sped up the search for the issue (which was with our codebase, not the library)

Due to the auto-exporter not exporting private members (`_matches`) it's the difference between inspecting `responses.matches` and `responses.mock._matches` during debugging. Not a huge issue once you know about it but not very accessible and not easy to find for people just starting out with the lib that might run into issues. If you'd prefer not to have the public method and just update the docs I can do that also.

markstory: For now I think updating the docs is ok. If this situation comes up more we can expand the public interface.

PR #254, opened 2019-01-29
Clean up 'mock' requirement declaration in setup.py

jdufresne:
For Python 2.7, mock is a hard requirement and so should be listed in `install_requires`.

Update system requirement to the more accurate: `python_version < '3.3'`. It no longer references unsupported Python 2.6.

Alphabetize the install_requires list.


markstory:
Thank you 👍

PR #259, opened 2019-02-24
Issue 225: Connection Error message mentions Responses

### LINKED ISSUE #225: Update ConnectionError message to mention that the error is from Responses
Similar to https://github.com/getsentry/responses/issues/77#issue-97890919, I had a confusing debugging session coming into a project that was using Responses. I was getting ConnectionError messages and found it difficult to work out what was wrong. It could be really useful to provide a custom error message in the ConnectionError that says something like "ConnectionError: tried to request a URL that doesn't have a Responses mock".

I'm not really familiar with the library, so I'm not sure if this exact message makes sense, but I think something like this could be quite helpful.

iamrpc:
Fixes #225 
Changed the message to:
1. show that the Connection was refused by Responses
2. Show that METHOD URL doesn't match with Responses Mock

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -589,8 +589,9 @@ class RequestsMock(object):
                 logger.info("request.allowed-passthru", extra={"url": request.url})
                 return _real_send(adapter, request, **kwargs)
 
-            error_msg = "Connection refused: {0} {1}".format(
-                request.method, request.url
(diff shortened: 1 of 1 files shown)


dcramer:
Solid improvement, thanks!

PR #262, opened 2019-03-15
Remove pythonPath key

arnavb:
Not everyone keeps their virtualenv in the location specified.


dcramer:
@arnavb this is how Sentry's development processes work - thats fine if you dont keep it there, but we do and this is catering towards us. It doesn't affect public releases.

dcramer:
(you'll notice this is in a lot of our repos: https://github.com/getsentry/sentry/blob/master/.vscode/settings.json)

arnavb:
Gotcha. Thanks for clarifying.

PR #268, opened 2019-05-11
Allow extra options to the `add_callback` API

### LINKED ISSUE #228: Allow to specify stream=True for callback responses
According to https://github.com/getsentry/responses/blob/master/responses.py#L355 default stream value for callback responses is False.
Also there is no way to override this value using public method add_callback https://github.com/getsentry/responses/blob/master/responses.py#L510
As a result this doesn't work:
```
        def callback(request):
            return 200, {}, 'content'
        responses.add_callback(
            'GET',
            'https://domain.com/',
            callback
        )
        response = requests.get('https://domain.com/', stream=True)
        content = response.raw.read()
....
    response.raw.read()
  File "/usr/lib/python3.6/site-packages/urllib3/response.py", line 380, in read
    data = self._fp.read()
ValueError: I/O operation on closed file.
```
Looks like it can be fixed by adding stream argument to add_callback method

As an workaround one can add a match manually using context manager:
```
        def callback(request):
            return 200, {}, 'content'
        with responses.RequestsMock() as requests_mock:

            requests_mock._matches.append(responses.CallbackResponse(
                'GET',
                'https://domain.com/',
                callback, stream=True
            ))
            response = requests.get('https://domain.com/', stream=True)
            content = response.raw.read()
```

krconv:
Adds and passes `kwargs` through to the underlying `CallbackResponse` object.

Closes #228 
Doing the same as #241 but includes a test

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -543,9 +543,20 @@ class RequestsMock(object):
         content_type="text/plain",
         **kwargs
     ):
-        # ensure the url has a default path set if the url is a string
-        # url = _ensure_url_default_path(url, match_querystring)
+        """
+        Register a callback to provide a dynamic response. The callback should
+        return a tuple of ``(status, headers, body)``.
 
+        >>> responses.add_callback(responses.GET, 'http://example.com', request_callback)
+
+        You can also pass a compiled regex to match multiple URLs.
+
+        >>> responses.add_callback(
+        >>>     responses.POST,
+        >>>     re.compile('http://calc.com/(sum|prod|pow|unsupported)'),
+        >>>     callback=request_callback,
+        >>> )
+        """
         self._matches.append(
             CallbackResponse(
                 url=url,
--- a/responses.py
+++ b/responses.py
@@ -534,7 +534,14 @@ class RequestsMock(object):
         self._matches[index] = response
 
     def add_callback(
-        self, method, url, callback, match_querystring=False, content_type="text/plain"
+        self,
+        method,
+        url,
+        callback,
+        match_querystring=False,
+        stream=False,
+        content_type="text/plain",
+        **kwargs
     ):
         # ensure the url has a default path set if the url is a string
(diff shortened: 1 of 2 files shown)


responses.py
```diff
@@ -534,18 +534,38 @@ def replace(self, method_or_response=None, url=None, body="", *args, **kwargs):
         self._matches[index] = response
 
     def add_callback(
-        self, method, url, callback, match_querystring=False, content_type="text/plain"
+        self,
+        method,
+        url,
+        callback,
+        match_querystring=False,
+        stream=False,
```
markstory: This parameter should come after `content_type`. People *could* be using positional arguments for `content_type` and adding a new parameter in the middle will break that use-case.

krconv: Thanks, I didn't consider that; I'll change that back

xzkostyan:
Any chance this PR could be merged?

markstory:
Do we have use cases for supporting forwarding `kwargs`. Doing so limits our options in the future, which is one reason I'm reluctant to merge this as is.

bblanchon:
`add_callback()` seems to be syntactic sugar on top of `CallbackResponse`, so I think it makes sense to forward all parameters, especially `stream`.

Any chances you guys can find a compromise and merge this?

markstory:
@bblanchon Is the `stream` option the one you're interested in? I'm :+1: for forwarding explicit parameters, am worried that including all kwargs limits our options going forward.

bblanchon:
Yes, I'm only interested in `stream`.

beliaev-maksim:
@markstory 
please close this PR as changes are made via #420

xzkostyan:
@beliaev-maksim does #420 allow `stream` parameter passing to `add_callback`?

beliaev-maksim:
@xzkostyan 
`stream` is deprecated. It has no influence if set on `Response` object. 

Set `stream` argument directly on request

PR #269, opened 2019-05-31
Return quoted value consistently

okomestudio:
This PR normalizes the return value of `_cookies_from_headers()` so that the function consistently returns URL-quoted version of non-ASCII characters set as cookie values.

The issue is that when a cookie value involves non-ASCII chars being URL-quoted, `cookies.Cookies.from_request()` appears to return unquoted chars whereas `http.cookies.SimpleCookie.load` does not. This inconsistency makes it tedious to write tests on cookies with non-ASCII characters.

### DIFF
--- a/test_responses.py
+++ b/test_responses.py
@@ -922,3 +922,18 @@ def test_custom_target(monkeypatch):
     requests_mock.start()
     assert len(patch_mock.call_args_list) == 1
     assert patch_mock.call_args[1]["target"] == "something.else"
+
+
+def _quote(s):
+    return responses.quote(responses._ensure_str(s))
+
+
+def test_cookies_from_headers():
+    text = "こんにちは/世界"
+    quoted_text = _quote(text)
+    expected = {"x": "a", "y": quoted_text}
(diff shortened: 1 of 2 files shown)


test_responses.py
```diff
@@ -922,3 +922,18 @@ def test_custom_target(monkeypatch):
...
+
+
+def test_cookies_from_headers():
+    text = "こんにちは/世界"
+    quoted_text = _quote(text)
+    expected = {"x": "a", "y": quoted_text}
+    headers = {"set-cookie": "; ".join(k + "=" + v for k, v in expected.items())}
+    cookiejar = responses._cookies_from_headers(headers)
+    for k, v in cookiejar.items():
+        assert isinstance(v, str)
```
markstory: Do you want `str` instances in python2?

okomestudio: That's the kind of design decision that I'm not exactly sure of, especially given that `__future__.unicode_literals` is  imported in the module.

My thinking is that `str` is more compatible with how cookies works in `requests` in Python 2.7, which simply store them as `str`, and that makes sense given that they contain URL-quoted chars, not non-ASCII chars, and for that the type doesn't need to be `unicode`.

PR #281, opened 2019-10-09
Fix testcase test_assert_all_requests_are_fired failure

litcoder:
Pytest 5.0.0 changed ExceptionInfo object's str() to returns the same as repr().
(See, item #5412 of https://docs.pytest.org/en/latest/changelog.html#pytest-5-0-0-2019-06-28)
This patch fixes the test failure by comparing expectation with str() result of
ExceptionInfo's value member instead of object itself.

Signed-off-by: Brandon Hong <brandon.hong@intel.com>


markstory:
Thank you 👍

PR #284, opened 2019-11-14
Fix tests on Python2.7

retracile:
pytest after 4.6 does not support Python 3.5 and earlier, causing tests run under Python 2.7 to fail with syntax errors in pytest.  Update the pytest requirement to use the appropriate version depending on the version of Python in use.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -27,7 +27,8 @@ install_requires = [
 ]
 
 tests_require = [
-    "pytest",
+    "pytest==4.6; python_version < '3.5'",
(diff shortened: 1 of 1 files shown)


markstory:
Thanks :+1:

StephenBrown2:
Pytest point releases are still being made on the 4.6 branch (currently at 4.6.6) until January 2020, so one should allow for that.
```py
pytest>=4.6,<5.0; python_version < '3.5'
pytest; python_version >= '3.5'
```

PR #273, opened 2019-07-01
Handle cookie attributes correctly (issue #120)

kathawala:
#120 

Cookies added through `adding-headers` would not be parsed correctly if they included `;` characters. `;` characters are treated as separators of cookies, rather than separators of the cookie attributes.

Previously:

```
Python 3.7.3 (default, Jun 24 2019, 04:54:02) 
[GCC 9.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import requests, responses
>>> @responses.activate
... def test():
...     responses.add(responses.GET, 'http://stackoverflow.com', adding_headers={'Set-Cookie': 'prov=c83dc7dc-e0b9-xxxx-xxxx-xxxxxxxxxxxx; domain=.stackoverflow.com; expires=Fri
, 01-Jan-2055 00:00:00 GMT; path=/; HttpOnly'})
...     return requests.get('http://stackoverflow.com').cookies
... 
>>> cookies = test()
>>> cookies
<RequestsCookieJar[Cookie(version=0, name='prov', value='c83dc7dc-e0b9-xxxx-xxxx-xxxxxxxxxxxx', port=None, port_specified=False, domain='', domain_specified=False, domain_initial_dot=False, path='/', path_specified=True, secure=False, expires=None, discard=True, comment=None, comment_url=None, rest={'HttpOnly': None}, rfc2109=False)]>
>>>
```

After this PR:

```
Python 3.7.3 (default, Jun 24 2019, 04:54:02) 
[GCC 9.1.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import requests, responses
>>> @responses.activate
... def test():
...     responses.add(responses.GET, 'http://stackoverflow.com', adding_headers={'Set-Cookie': 'prov=c83dc7dc-e0b9-xxxx-xxxx-xxxxxxxxxxxx; domain=.stackoverflow.com; expires=Fri
, 01-Jan-2055 00:00:00 GMT; path=/; HttpOnly'})
...     return requests.get('http://stackoverflow.com').cookies
... 
>>> cookies = test()
>>> cookies
[...]

PR #285, opened 2019-11-14
Fix cookie handling

retracile:
Cookie handling in responses is currently broken.  There have been a couple of PRs created to attempt to address this (PR #110, PR #273), but they do not handle `Set-Cookie` headers correctly.  In particular, a `Set-Cookie: session_id=12345; a=b; c=d` header should only create a `session_id` cookie, and _not_ create `a` or `c` cookies.  Multiple cookies are created using multiple `Set-Cookie` header lines, which means that the existing dict-based implementation cannot handle that.

This PR fixes the response objects to contain the additional members that the `requests` module relies on for cookie handling so that `responses` doesn't have to deal with it explicitly, uses `HTTPHeaderDict` to support multiple `Set-Cookie` headers, updates the existing cookie testcase to expect the correct behavior, and adds a couple more testcases to further demonstrate correct functionality.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -23,6 +23,10 @@ try:
     from requests.packages.urllib3.response import HTTPResponse
 except ImportError:
     from urllib3.response import HTTPResponse
+try:
+    from requests.packages.urllib3.connection import HTTPHeaderDict
+except ImportError:
+    from urllib3.connection import HTTPHeaderDict
 
 if six.PY2:
     from urlparse import urlparse, parse_qsl, urlsplit, urlunsplit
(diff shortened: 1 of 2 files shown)


retracile:
I would like to get the cookie handling fixed; any feedback on this PR that I should address?

retracile:
@markstory Is there anything I can provide to improve your comfort level with this PR?

markstory:
@retracile No, I've just been swamped the last few weeks and haven't been able to test this against an application that uses cookies + responses. I'm hoping this week will be better though.

retracile:
@markstory Ah, ok.  Thanks for the update.

VeNoMouS:
Can this get merged already!?

VeNoMouS:
<3 @markstory thanks dude

retracile:
@markstory Thanks!

This fix was included in the 0.10.8 release.

PR #298, opened 2020-01-18
Add ability to pass a regex to add_passthru.

### LINKED ISSUE #264: ability to pass a regex to add_passthru
e.g:
`responses.add_passthru(re.compile(r'http(s?):\/\/(\w+)(:*)(\d*)\/api\/v1\/activities\/'))` 
so that calls to any domain[:port] with this request path may call the upstream server.

hramezani:
Fix #264

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -527,14 +527,18 @@ class RequestsMock(object):
 
     def add_passthru(self, prefix):
         """
-        Register a URL prefix to passthru any non-matching mock requests to.
+        Register a URL prefix or regex to passthru any non-matching mock requests to.
 
         For example, to allow any request to 'https://example.com', but require
         mocks for the remainder, you would add the prefix as so:
 
         >>> responses.add_passthru('https://example.com')
+
+        Regex can be used like:
+
+        >>> responses.add_passthru(re.compile('https://example.com/\\w+'))
         """
-        if _has_unicode(prefix):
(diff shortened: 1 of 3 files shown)


responses.py
```diff
@@ -527,16 +527,20 @@ def add(
 
     def add_passthru(self, prefix):
         """
-        Register a URL prefix to passthru any non-matching mock requests to.
+        Register a URL regex to passthru any non-matching mock requests to.
```
markstory: Isn't it a prefix or regex now?

responses.py
```diff
@@ -527,16 +527,20 @@ def add(
...
         >>> responses.add_passthru('https://example.com')
+
+        Regex can be used like:
+
+        >>> responses.add_passthru('https://example.com/\\w+')
         """
         if _has_unicode(prefix):
             prefix = _clean_unicode(prefix)
-        self.passthru_prefixes += (prefix,)
+        self.passthru_prefixes += (re.compile(prefix),)
```
markstory: This seems like it could fail in scenarios where the pass through prefix contains re metacharacters. Having an re pattern passed in as a parameter avoids backwards compatibility issues and makes the user's intent very clear, as str and re patterns are different types.

hramezani: Yes, you are right.
I've changed the PR with your idea.

PR #299, opened 2020-01-19
Add how to run tests to readme.

hramezani:


README.rst
```diff
@@ -422,3 +422,15 @@ Configure development requirements:
 .. code-block:: shell
 
     make develop
+
+Run all tests:
+
+.. code-block:: shell
+
+    pytest .
```
dcramer: It's just 'py.test'. I'm not sure the README needs to explain "how to pytest" though, but I think its fair to say that it uses pytest.

hramezani: I have updated the PR:

- Mentioned that Responses uses `Pytest`
- Fixed the path

I think it is common to explain `how to run the tests` in the `Contributing` section of `README`.

PR #300, opened 2020-01-19
Add for BaseResponse its own call list

raizanshine:
Responses has only common call list for requests.
I lack ability to retrieve call list for specific match when I write tests for complex internal API which uses `requests` implicitly.

For example:
```python
def test_something():
    responses.add(
        responses.GET, 'http://example.com/first/,
        json={'success': 'true'}
    )
    responses.add(
        responses.GET, 'http://example.com/second/,
        json={'success': 'false'}, status_code=400
    )
    # calling method which revokes requests
    call_internal_api()
    # forced to get the calls from common call list
    first_call, second_call = responses.calls
```

This is inconvinient because I have to keep in mind the order of the calls and assign all of them (even if I don't need them all). Or I need to extract calls with filtering by request url.

I had a quick look at issues and PRs but didn't found something similar.
I hope this could be useful for someone else.


test_responses.py
```diff
@@ -990,3 +990,76 @@ def test_cookies_from_headers():
...
         assert isinstance(v, str)
         assert v == expected[k]
+
+
+def test_response_match_has_own_calls():
+    @responses.activate
+    def run():
+        match = responses.add(method=responses.GET, url="http://example.com", body="OK")
+        requests.get("http://example.com")
+        resp = match.calls[0].response
```
markstory: It would be better if this test created two responses and called both of them. That would help ensure that each stub tracked responses separately. Right now that _could_ be true, but with no other stub we can't be sure.

raizanshine: Test was extended 6b085bd

beliaev-maksim:
@markstory 
I think having a `CallList` on the `BaseResponse` is a bit confusing. Currently we apply `calls` only on `RequestMock`, so, you can check all calls (request-response pair) that are done.

applying `CallList` on a response will be a redundant addition. Basically, response referring to itself + request object


considering no votes in the last couple of years, I propose to close incomplete

PR #301, opened 2020-01-19
Return requested params in responses.calls[].request.

### LINKED ISSUE #271: Feature Request: Return requested params in responses.calls[].request
Similar to as discussed in #148 , when using the params parameter in requests, I need to set `match_querystring=False` to ignore the query string:

```
responses.add(
    method=responses.GET,
    url="http://example.com?hello=world,
    json={"response": "response"},
    status=200,
    match_querystring=False,
)
req = requests.get('http://example.com', params={"hello": "world"})
```

This returns a list of calls where I can get the requested URL:
```
print(requests.calls[0].request.url)
> http://example.com
```

However, theres no way to see what the original params passed in the original `request`. It would be nice to be able to also do something like the following so that I can very what the parameters are that were passed in the request:

```
print(requests.calls[0].request.params)
> {"hello": "world"}
```

hramezani:
Fix #271

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -239,6 +239,25 @@ a callback function to give a slightly different result, you can use ``functools
         )
 
 
+You can see params passed in the original ‍‍``request`` in ‍‍``responses.calls[].request.params``:
+
+.. code-block:: python
+
+    import responses
+    import requests
+
+    @responses.activate
+    def test_request_params():
+        responses.add(
+            method=responses.GET,
+            url="http://example.com?hello=world",
+            body="test",
+            match_querystring=False,
+        )
+
+        resp = requests.get('http://example.com', params={"hello": "world"})
+        assert response.calls[0].request.params == {"hello": "world"}
+
 Responses as a context manager
 ------------------------------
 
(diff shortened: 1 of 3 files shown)


responses.py
```diff
@@ -622,6 +622,7 @@ def _find_match(self, request):
     def _on_request(self, adapter, request, **kwargs):
         match = self._find_match(request)
         resp_callback = self.response_callback
+        request.params = dict(parse_qsl(request.path_url.replace("/?", "")))
```
markstory: Could we include this in the example under Dynamic Responses?

hramezani: Example has been added.

README.rst
```diff
@@ -239,6 +239,25 @@ a callback function to give a slightly different result, you can use ``functools
         )
 
 
+You can see params passed in the original ‍‍``request`` in ‍‍``responses.calls[].request.paramas``:
```
markstory:
```suggestion
You can see params passed in the original ‍‍``request`` in ‍‍``responses.calls[].request.params``:
```

PR #302, opened 2020-01-24
Change RequestsMock._on_request to catch BaseException.

### LINKED ISSUE #297: RequestsMock._on_request misses calls if callback raises BaseException
In case a `BaseException` (or a subclass) is raised in the `get_response()` method of a match object (e.g. in a callback) the call accounting isn't updated for that call.

This then leads to test failures due to `Not all requests have been executed`.

The reason is that only `Exception`s are expected:
https://github.com/getsentry/responses/blob/8b0b9c1883c1242246af3dd535df79bfb071732a/responses.py#L642-L648

I noticed this in the context of `gevent` and its `Timeout` feature which raises an exception inherited from `BaseException`.

hramezani:
Fix #297

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -652,7 +652,7 @@ class RequestsMock(object):
 
         try:
             response = adapter.build_response(request, match.get_response(request))
-        except Exception as response:
+        except BaseException as response:
             match.call_count += 1
             self._calls.add(request, response)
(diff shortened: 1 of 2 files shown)


markstory:
Thank you 👍

Issue #306, opened 2020-02-24
Regression between 0.10.7 and 0.10.8


kvdveer:
Minimal reproduction:
```
import responses
import requests


@responses.activate
def test_it():
    
    def request_callback(request):
        return (
            200,
            [('Content-Type', 'application/json')],
            b'foo',
        )

    responses.add_callback('GET', 'http://mockhost/.foo', callback=request_callback)
    resp = requests.get('http://mockhost/.foo')
    assert resp.headers['content-type'] == 'application/json'
```

This test succeeds on 0.10.7, and fails on 0.10.8.

This may be the root cause behind #292, but that ticket contains insufficient details to confirm this suspicion, so I created a new ticket instead.

PR #308, opened 2020-02-25
fix: Fix content-type in callbacks using list headers

markstory:
Fix the content type header in list based callback headers.

Refs #306

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -408,7 +408,14 @@ class CallbackResponse(BaseResponse):
         # If the callback set a content-type remove the one
         # set in add_callback() so that we don't have multiple
         # content type values.
-        if "Content-Type" in r_headers:
+        has_content_type = False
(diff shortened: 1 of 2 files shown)


responses.py
```diff
@@ -408,7 +408,14 @@ def get_response(self, request):
         # If the callback set a content-type remove the one
         # set in add_callback() so that we don't have multiple
         # content type values.
-        if "Content-Type" in r_headers:
+        has_content_type = False
+        if isinstance(r_headers, dict) and "Content-Type" in r_headers:
+            has_content_type = True
+        elif isinstance(r_headers, list):
+            has_content_type = any(
+                [h for h in r_headers if len(h) and h[0].lower() == "content-type"]
```
hramezani: I think you can replace `len(h)` with `h`:
```suggestion
                [h for h in r_headers if h and h[0].lower() == "content-type"]
```

PR #311, opened 2020-03-27
Stop deleting a callbackResponse when there is more than one match with pattern

alanoe:
Registered callbackResponses should never be deleted automatically.


markstory:
If the matcher is not removed how would callbacks matching the same pattern/prefix ever be reached?

alanoe:
> If the matcher is not removed how would callbacks matching the same pattern/prefix ever be reached?

Do they need to be reached? Calling only the callback associated with the first matching regex is  the most usual behavior in regular expression matchers (eg Django URL matcher). 

As this is the most common behavior, clients often implement inverse matching using a specific rule followed by a less specific one, instead of using regex negative look-around.  Example found in the moto library which is breaking my app test code and motivated me to create this PR (https://github.com/spulec/moto/blob/acfdfb7d7c08be05a02101ff10cf467015bb0158/moto/core/models.py#L338 ):

regex rules:
1) https?://s3(.*).amazonaws.com/$ -> mock it
2)https?://s3(.*).amazonaws.com/(?P<key_or_bucket_name>[^/]+)/?$ -> mock it
3)https?://(?P<bucket_name>[a-zA-Z0-9\-_.]*)\.?s3(.*).amazonaws.com/(?P<key_or_bucket_name>[^/]+)/?$  -> mock it
4)https?://s3(.*).amazonaws.com//(?P<key_name>.*)$
5) https?://(?P<bucket_name>[a-zA-Z0-9\-_.]*)\.?s3(.*).amazonaws.com/$
6) https?://(?P<bucket_name>[a-zA-Z0-9\-_.]*)\.?s3(.*).amazonaws.com//(?P<key_name>.*)$
7) https?://(?P<bucket_name>[a-zA-Z0-9\-_.]*)\.?s3(.*).amazonaws.com/(?P<bucket_name_path>[^/]+)/(?P<key_name>.+)
8) https?://.+.amazonaws.com/.* -> return HTTP response with status code 400

If I send 3 HTTP requests with the URL https://s3.amazonaws.com/mybucket/mykey , the first one will be matched with rule 3, the second one with rule 7 and all subsequent requests will be matched with rule 8 and the mock will fail.

Anyway, IMHO there are 2 implementations which make sense 
- for each HTTP request, 
- for each HTTP request, call all the callbacks associated with a matching regex

However, the current implementation in this library calls the CallbackResponse associated with the first matching regex *on each HTTP request* and this regex is deleted from the list of registered regexes. This causes 2 inconsistencies:
- if same URL is accessed twice, different CallbackResponses will be called
- preference is given to the callback associated with the matching regex which was registered last. For instance, if there are 2 regexes which match a URL, only the first HTTP request will call the cbResponse associated with the first registered regex and all subsequent calls will call the cbResponse associated with the second registered regex.

markstory:
Not all callback responses use regular expression patterns to match URLs though. Some are just fixed strings that do substring matching. The tests that are failing cover the scenario I'm concerned about which would be broken by these changes.

alanoe:
> Not all callback responses use regular expression patterns to match URLs though. Some are just fixed strings that do substring matching. The tests that are failing cover the scenario I'm concerned about which would be broken by these changes.

I have checked one of the tests (test_multiple_responses() ) to try to understand the usefulness of matching the same url/url substring to different responses, but I was not really able to. It seems to me the client could simply use a single callback method and return the different responses he wanted, in any order. Is this feature implemented just for convenience?

Anyway, the current behavior which sometimes keeps the matching rules (regardless if it is expressed by a regex or substring) is inconsistent. If there are 2 callbacks for the same match, callbacks are incrementally deleted. If there are not, they are preserved. IMHO that should be fixed.

Issue #313, opened 2020-04-06
Feature request: Asserting calls with POST parameters


samrussell:
Use case: testing an app makes an API call with the correct parameters

Currently we can get some of the way as follows:

```
import responses
import requests

def test_my_api():
    with responses.RequestsMock(assert_all_requests_are_fired=True) as rsps:
        rsps.add(responses.GET, 'http://twitter.com/api/1/foobar?asdf=zcxv',
                 match_querystring=True,
                 body='{}', status=200,
                 content_type='application/json')
```

This will raise an exception at the end of the context if the route `http://twitter.com/api/1/foobar` hasn't been hit with the query string `asdf=zcxv`. I'd like to extend this so we can expect certain form parameters, e.g.:

```
import responses
import requests

def test_my_api():
    with responses.RequestsMock(assert_all_requests_are_fired=True) as rsps:
        rsps.add(responses.GET, 'http://twitter.com/api/1/foobar',
                 request_content_type='application/x-www-form-urlencoded',
                 params={'asdf': 'zcxv'}
                 body='{"custom": "body"}', status=200,
                 content_type='application/json')
```

beliaev-maksim:
@samrussell 
can you check if new `matchers` solve your problem?

samrussell:
If you're referring to #319 then yes, it solves my problem perfectly

beliaev-maksim:
@samrussell
Great, than you can consider closing this issue

PR #314, opened 2020-04-07
Allow requests to match on POSTed parameters

samrussell:
Implements #313

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -235,11 +235,12 @@ class BaseResponse(object):
 
     stream = False
 
-    def __init__(self, method, url, match_querystring=_unspecified):
+    def __init__(self, method, url, match_querystring=_unspecified, post_params=None):
         self.method = method
         # ensure the url has a default path set if the url is a string
         self.url = _ensure_url_default_path(url)
         self.match_querystring = self._should_match_querystring(match_querystring)
+        self.post_params = post_params
         self.call_count = 0
 
     def __eq__(self, other):
@@ -312,6 +313,12 @@ class BaseResponse(object):
         else:
             return False
 
+    def _post_params_match(self, post_params, request_body, content_type):
+        if post_params == None:
+            return True
+
+        return sorted(post_params.items()) == sorted(parse_qsl(request_body))
+
     def get_headers(self):
         headers = HTTPHeaderDict()  # Duplicate headers are legal
         if self.content_type is not None:
@@ -328,6 +335,9 @@ class BaseResponse(object):
             return False
 
         if not self._url_matches(self.url, request.url, self.match_querystring):
+            return False
+
+        if not self._post_params_match(self.post_params, request.body, self.content_type):
             return False
 
         return True
--- a/responses.py
+++ b/responses.py
@@ -314,7 +314,7 @@ class BaseResponse(object):
             return False
 
     def _post_params_match(self, post_params, request_body, content_type):
-        if post_params == None:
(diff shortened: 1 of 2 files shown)


samrussell:
What's the deal with .5 and .6? The test runs seemed to be fine but a pre-commit task is failing?

responses.py
```diff
@@ -312,6 +313,12 @@ def _url_matches(self, url, other, match_querystring=False):
         else:
             return False
 
+    def _post_params_match(self, post_params, request_body, content_type):
+        if post_params == None:
```
kathawala, line 317: To satisfy the linter, this should be changed to

```python
if post_params is None:
```

samrussell: nice catch, thank you! need more than a falsey comparison ({} would be valid to match on, for example)

samrussell: what's a good way to run the linter locally? is it pylint or similar?

kathawala: I think the linter used is [https://github.com/psf/black](url) if you want to run locally

responses.py
```diff
@@ -312,6 +313,12 @@ def _url_matches(self, url, other, match_querystring=False):
         else:
             return False
 
+    def _post_params_match(self, post_params, request_body, content_type):
+        if post_params is None:
+            return True
+
+        return sorted(post_params.items()) == sorted(parse_qsl(request_body))
```
markstory, line 320: How would this work for non URL encoded post bodies?

samrussell: That's a very good point :) I have `x-www-form-urlencoded` in mind but obviously would need to check for that and exclude others.

The use case I'm looking at is testing a Slack integration to make sure it passes the right params upstream. It gets called with a list of users to invite, and I want to confirm that it's POSTing the correct IDs. The pattern looks something like this:

```
with responses.RequestsMock() as request_mock:
  request_mock.add(responses.POST, <matcher that matches the URL and POSTed params>, body=blah, status=200)
  slack_api.invite_users(["jonny", "mark", "lisa"])
```

My version only works with `x-www-form-urlencoded`, but then other payloads vary... Would it be better to allow responses/RequestsMock.add to take some sort of Matcher object that looks at the URL and then looks at the request headers/body to parse out params?

samrussell: I've done a version that take custom matchers and added implementations for both JSON and urlencoding on #319. I'll have a go at a different version of post_params that uses the content-type to decide how to match

samrussell: I got it working with the `post_params` argument, and there's code now that uses `Content-Type` to decide how to decode and compare the body - PR #320

samrussell: Hi, have you had a chance to take another look at this? I think either #319 or #320 would both be good things that fit with the project, but if not I'm happy to try another tack

markstory: I talked with some of the other folks at sentry and an approach similar to #319 is more along the lines of what we would prefer. Having a flexible single keyword argument that allows multiple matchers to be used is preferable over embedding implicit content-type logic or adding more parameters for each request dimension.

If its ok with you, I'd like to close this pull request and #320 and focus on #319

samrussell:
That's fine by me, please let me know anything #319 is missing and I'm
happy to do some more work there

RochesterinNYC:
Just found this feature/issue and noting that I/we would find the ability to do this in `responses` very useful as well.

markstory:
Closing in favour of #319

PR #315, opened 2020-04-07
Improve code coverage

samrussell:
Was looking over the code coverage report and added some missing test cases:

* `BaseResponse.matches()` distinguishing on method
* `BaseResponse._url_matches_strict()` when the number of params is the same but the keys don't match
* `_handle_body()` when the body is a BufferedReader from opening a file handle

The third commit here actually simplifies the comparison in `BaseResponse._url_matches_strict()` to a straight list comparison - it doesn't make sense to compare length and then directly compare the contents of the tuples, unless this is leftover from some old dict comparison code?

### DIFF
--- a/test_responses.py
+++ b/test_responses.py
@@ -334,6 +334,20 @@ def test_match_querystring_auto_activates():
     assert_reset()
 
 
+def test_match_querystring_missing_key():
+    @responses.activate
+    def run():
+        responses.add(responses.GET, "http://example.com?foo=1&bar=2", body=b"test")
+        with pytest.raises(ConnectionError):
+            requests.get("http://example.com/?foo=1&baz=2")
+
+        with pytest.raises(ConnectionError):
+            requests.get("http://example.com/?bar=2&fez=1")
+
+    run()
+    assert_reset()
+
+
 def test_accept_string_body():
     @responses.activate
     def run():
--- a/test_responses.py
+++ b/test_responses.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import, print_function, division, unicode_litera
 import inspect
 import re
 import six
+from io import BufferedReader, BytesIO
 
 import pytest
 import requests
@@ -914,6 +915,21 @@ def test_handles_unicode_url():
     @responses.activate
     def run():
         responses.add(responses.GET, url, body="test")
+
+        resp = requests.get(url)
+
+        assert_response(resp, "test")
+
+    run()
+    assert_reset()
+
+
+def test_handles_buffered_reader_body():
+    url = "http://example.com/test"
+
+    @responses.activate
+    def run():
+        responses.add(responses.GET, url, body=BufferedReader(BytesIO(b"test")))
 
         resp = requests.get(url)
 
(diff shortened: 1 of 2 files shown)


test_responses.py
```diff
@@ -334,6 +335,20 @@ def run():
...
+    def run():
+        responses.add(responses.GET, "http://example.com?foo=1&bar=2", body=b"test")
+        with pytest.raises(ConnectionError):
+            requests.get("http://example.com/?foo=1&baz=2")
+
+        with pytest.raises(ConnectionError):
+            requests.get("http://example.com/?bar=2&fez=1")
+
+    run()
+    assert_reset()
```
hramezani, line 349: I think it would be good to have a test for query string params count like: `http://example.com/?foo=1`.
Also, you can use [pytest parametrize](https://docs.pytest.org/en/latest/parametrize.html).
```suggestion
@pytest.mark.parametrize(
    "test_url",
    [
        "http://example.com/?foo=1&baz=2",
        "http://example.com/?bar=2&fez=1",
        "http://example.com/?foo=1",
    ]
)
def test_match_querystring_missing_key(test_url):
    @responses.activate
    def run():
        responses.add(responses.GET, "http://example.com?foo=1&bar=2", body=b"test")
        with pytest.raises(ConnectionError):
            requests.get(test_url)

    run()
    assert_reset()

```

samrussell: This is already tested elsewhere. The goal of this commit was to handle the case where the keys don't match and everything else does; ultimately changing the code to a straight comparison between the two lists removes the need for this edge case anyway.

test_responses.py
```diff
@@ -909,6 +924,21 @@ def run():
...
 
+def test_handles_buffered_reader_body():
+    url = "http://example.com/test"
+
+    @responses.activate
+    def run():
+        responses.add(responses.GET, url, body=BufferedReader(BytesIO(b"test")))
+
+        resp = requests.get(url)
+
```
hramezani, line 935: Please chop blank line.

samrussell: This looks identical to the other tests above:

```
def test_handles_unicode_querystring():
    url = "http://example.com/test?type=2&ie=utf8&query=汉字"

    @responses.activate
    def run():
        responses.add(responses.GET, url, body="test", match_querystring=True)

        resp = requests.get(url)

        assert_response(resp, "test")

    run()
    assert_reset()


def test_handles_unicode_url():
    url = "http://www.संजाल.भारत/hi/वेबसाइट-डिजाइन"

    @responses.activate
    def run():
        responses.add(responses.GET, url, body="test")

        resp = requests.get(url)

        assert_response(resp, "test")

    run()
    assert_reset()


def test_handles_buffered_reader_body():
    url = "http://example.com/test"

    @responses.activate
    def run():
        responses.add(responses.GET, url, body=BufferedReader(BytesIO(b"test")))

        resp = requests.get(url)

        assert_response(resp, "test")

    run()
    assert_reset()
```

I'd be breaking the convention of the file if I was to chop the blank line there.

markstory:
Thank you 👍

PR #316, opened 2020-04-07
Fix bug in handling unicode body

samrussell:
Added a test to return a unicode body and noticed this wasn't working as
the charset wasn't being set. This also highlighted that Response
converts unicode to bytes twice, once in the constructor and once in the
call to _handle_body(), so the conversion in the constructor has been
removed.

There's possibly a similar bug in CallbackResponse too.


markstory:
Thank you 👍

PR #317, opened 2020-04-10
Fix params parsing

jllorencetti:
Using a simple replace when parsing the params results in an error when there's some path present.
For example, without this fix the new test would break with something like:
```python
AssertionError:
    {'/some/path?hello': 'world', 'example': 'params'}
    {'example': 'params', 'hello': 'world'}
```

### DIFF
--- a/test_responses.py
+++ b/test_responses.py
@@ -1122,21 +1122,22 @@ def test_cookies_from_headers():
         assert v == expected[k]
 
 
-def test_request_param():
+@pytest.mark.parametrize("url", ("http://example.com", "http://example.com/some/path",))
+def test_request_param(url):
     @responses.activate
     def run():
         params = {"hello": "world", "example": "params"}
         responses.add(
             method=responses.GET,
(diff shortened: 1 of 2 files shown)


test_responses.py
```diff
@@ -1122,21 +1122,22 @@ def test_cookies_from_headers():
         assert v == expected[k]
 
 
-def test_request_param():
+@pytest.mark.parametrize("url", ("http://example.com", "http://example.com/some/path",))
```
markstory, line 1125: It would be good to have a URL with a trailing `/` in here to.

jllorencetti:
@markstory done. Don't know why the checks are still pending though.

georgeyk:
Hey, any plans on merging this one?

jkaur26:
Hi. When will this change be released to Pypi?

markstory:
@jkaur26 Probably in the next few days. There is another pull request that is close to being ready for release.

jkaur26:
@markstory Thank you for the update.

PR #318, opened 2020-04-11
Added assert_called_once.

hramezani:


responses.py
```diff
@@ -692,6 +692,23 @@ def stop(self, allow_assert=True):
                 )
             )
 
+    def assert_called_once(self, url):
```
markstory, line 695: Would getting the call count for a URL give a more general purpose API? For example you'd be able to assert 0, or N calls with the same amount of code as 1.

hramezani: @markstory, Thanks for your suggestion.
I've changed it to accept `call_count`.
Please let me know if the function name or documentation needs changes.

responses.py
```diff
@@ -692,6 +692,23 @@ def stop(self, allow_assert=True):
                 )
             )
 
+    def assert_call_count(self, url, count):
+        call_count = len(
+            [
+                1
+                for call in self.calls
+                if call.request.url == _ensure_url_default_path(url)
```
markstory, line 700: Should we fail early when the request is empty?

responses.py
```diff
@@ -692,6 +692,23 @@ def stop(self, allow_assert=True):
...
+            ]
+        )
+        if call_count == count:
+            return True
+        else:
+            raise AssertionError(
+                "Expected request to be called {0} times. Called {1} times.".format(
```
markstory, line 707: Could this include the URL that was provided? It could be helpful in figuring out which test is failing if assertions are made in a loop.

hramezani: URL included.

responses.py
hramezani: I don't know how I can test it `when the request is empty`.(TBH I didn't understand it).

FYI, we can use this for testing 0 calls.
https://github.com/getsentry/responses/pull/318/files#diff-a95e1a562f3879ddb72fe81003f48c52R1164

Please explain the case, I will change the patch.
Thanks

PR #319, opened 2020-04-12
Add post params with custom matchers

samrussell:
This approach lets users match on POSTed parameters - it includes two matchers for urlencoded and JSON encoded parameters and leaves open the possibility of matching on params encoded in other ways:

```
responses.add(
    method=responses.POST,
    url="http://example.com/",
    body="one",
    post_params_matcher=responses.json_params_matcher(
        {"page": {"name": "first", "type": "json",}}
    ),
)
responses.add(
    method=responses.POST,
    url="http://example.com/",
    body="two",
    post_params_matcher=responses.urlencoded_params_matcher(
        {"page": "second", "type": "urlencoded"}
    ),
)
```

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -247,7 +247,7 @@ def json_params_matcher(params):
             if isinstance(request_body, bytes):
                 request_body = request_body.decode("utf-8")
             return params == json_module.loads(request_body)
-        except JSONDecodeError as err:
+        except JSONDecodeError:
             return False
 
     return match
--- a/responses.py
+++ b/responses.py
@@ -229,18 +229,37 @@ def _handle_body(body):
 _unspecified = object()
 
 
+def urlencoded_params_matcher(params):
+    def match(request_body):
+        return sorted(params.items()) == sorted(parse_qsl(request_body))
+
+    return match
+
+
+def json_params_matcher(params):
+    def match(request_body):
+        try:
+            return params == json_module.loads(request_body)
+        except json_module.decoder.JSONDecodeError as err:
+            return False
+
+    return match
+
+
 class BaseResponse(object):
     content_type = None
     headers = None
 
     stream = False
 
-    def __init__(self, method, url, match_querystring=_unspecified, post_params=None):
+    def __init__(
+        self, method, url, match_querystring=_unspecified, post_params_matcher=None
+    ):
         self.method = method
         # ensure the url has a default path set if the url is a string
         self.url = _ensure_url_default_path(url)
         self.match_querystring = self._should_match_querystring(match_querystring)
-        self.post_params = post_params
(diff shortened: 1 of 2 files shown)


markstory:
I like the direction you've taken here. I think the more extensible post data matcher is a great idea. Do you think we could extend it another step and generalize matchers more? For example

```python
# Match a json body
responses.add(
  responses.POST,
  url='http://example.com/api/people',
  body='oh no',
  match=[responses.match_json_body(a='b', c='d')]
)

# Match a urlencoded body
responses.add(
  responses.POST,
  url='http://example.com/api/people',
  body='oh no',
  match=[responses.match_urlencoded_body(a='b', c='d')]
)
```

This would enable userland code to add custom matchers, and furthermore we could refactor the internal url + content-type matching to use this same matcher flow.

samrussell:
I think I understand - we have a generic `match` parameter that takes some number of matchers, and that keeps the interface more open for expansion in future? I think I'll need to add some more test work around it, but the only change to the code is just renaming the parameter and making take a list of matchers instead of a single matcher, is that right?

samrussell:
I've rebased on master and done a minor change to add `match=[]` as a param.

The matchers I've added are mutually exclusive so I've not added coverage for multiple matchers. There could be some future work for partial matching, but this feature will give users to add their own customer matchers in any case.

responses.py
```diff
@@ -302,6 +327,13 @@ def _url_matches(self, url, other, match_querystring=False):
         else:
             return False
 
+    def _body_matches(self, match, request_body):
+        for matcher in match:
+            if not matcher(request_body):
```
markstory, line 332: I wonder if this should be `body` so that if in the future we add query string or URL the keyword arguments would have succinct names.

test_responses.py
```diff
@@ -1184,3 +1184,47 @@ def run():
...
+        )
+        responses.add(
+            method=responses.POST,
+            url="http://example.com/",
+            body="two",
+            match=[
+                responses.urlencoded_params_matcher(
+                    {"page": "second", "type": "urlencoded"}
+                )
+            ],
```
markstory, line 1210: I like how this interface turned out. It gives us a flexible path for all sorts of custom matchers to be implemented.

responses.py
samrussell: The problem we run into is that the Response has its own `body`. This method gets called like so:

```
        if not self._url_matches(self.url, request.url, self.match_querystring):
            return False

        if not self._body_matches(self.match, request.body):
            return False
```

So it makes sense to pass the body from the request in this way, but if it's just called `body` in the function then we end up in a bug-prone state where `body` and `self.body` are different things. Open to ideas on this.

PR #320, opened 2020-04-13
Add post params JSON or urlencoded

samrussell:
This allows matching on POSTed parameters by passing a normal python map, and this is parsed out when a request is received based on the Content-Type of the request


markstory:
Closing in favor of #319

PR #327, opened 2020-06-10
Added Py39 support

hramezani:
Python 3.9 release is on October 5, 2020.

### DIFF
--- a/test_responses.py
+++ b/test_responses.py
@@ -396,9 +396,9 @@ def test_arbitrary_status_code():
     @responses.activate
     def run():
         url = "http://example.com/"
-        responses.add(responses.GET, url, body="test", status=418)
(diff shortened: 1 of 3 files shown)


test_responses.py
```diff
@@ -399,7 +400,10 @@ def run():
         responses.add(responses.GET, url, body="test", status=418)
         resp = requests.get(url)
         assert resp.status_code == 418
-        assert resp.reason is None
```
mgorny, line 402: I'm not sure if the code is 'arbitrary' now, though. I suppose the purpose might be to use a code that isn't supported.

hramezani: Good point :+1: 
I changed the status code to `419`

markstory:
Thank you 👍

PR #329, opened 2020-06-15
Add custom reason

sac-cas:
Added reason to responses so any status code can be mocked with a given reason

### DIFF
--- a/test_responses.py
+++ b/test_responses.py
@@ -24,9 +24,11 @@ def assert_reset():
     assert len(responses.calls) == 0
 
 
-def assert_response(resp, body=None, content_type="text/plain"):
-    assert resp.status_code == 200
-    assert resp.reason == "OK"
+def assert_response(
+    resp, body=None, content_type="text/plain", status_code=200, reason="OK"
+):
+    assert resp.status_code == status_code
+    assert resp.reason == reason
     if content_type is not None:
         assert resp.headers["Content-Type"] == content_type
     else:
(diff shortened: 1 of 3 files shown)


test_responses.py
```diff
@@ -54,6 +56,34 @@ def run():
...
+    @responses.activate
+    def run():
+        responses.add(
+            responses.GET,
+            "http://example.com",
+            body=b"take care!",
+            status=299,
+            reason="warning",
+        )
+        requests.get("http://example.com")
```
hramezani, line 69: I think we need to just test the newly added feature. something like:

```python
        assert resp.status_code == 299
        assert resp.reason == "warning"

```

test_responses.py
```diff
@@ -54,6 +56,34 @@ def run():
...
+        assert len(responses.calls) == 1
+        assert responses.calls[0].request.url == "http://example.com/"
+        assert responses.calls[0].response.content == b"take care!"
+        assert responses.calls[0].response.reason == "warning"
+        assert responses.calls[0].response.status_code == 299
+
+        requests.get("http://example.com?foo=bar")
```
hramezani, line 76: What is the purpose of this test?

test_responses.py
```diff
@@ -127,6 +157,46 @@ def run():
...
+                method=responses.GET,
+                url="http://example.com/api",
+                body="replacement",
+                status=299,
+                reason="warning",
+            ),
+        ),
+    ],
+)
+def test_replace_with_reason(original, replacement):
```
hramezani, line 180: I think we don't need this test.

hramezani:
I don't know for sure if we want this or not. Other than that, we need to add [documentation to README.rst](https://github.com/getsentry/responses#response-parameters).

test_responses.py
sacastells: The test is similar to `test_response` just with custom status and reason. We can only test the newly added features. In my code I checked all values just to make an entire test.

test_responses.py
sacastells: Replace must be tested separately I think. The given `test_replace` adds more responses than necessary so I'm not sure what is its purpose.

test_responses.py
sacastells: The test is similar to `test_response` just with custom status and reason. We can only test the newly added features. In my code I checked all values just to make an entire test.

test_responses.py
sacastells: We can only test the newly added features. In my code I checked all values just to make an entire test. I think it's at least not worse to test the other values :)

test_responses.py
sacastells: Replace must be tested separately I think. The given test_replace adds more responses than necessary so I'm not sure what is its purpose.

sacastells:
I only added the reason text to add/replace not to callback btw.

markstory:
Is your usecase for this to enable status text on non-standard response codes? Or something else?

sacastells:
> Is your usecase for this to enable status text on non-standard response codes? Or something else?

Yes, my only usecase is to enable custom status text for standard response codes. It can also be used for standard usecases but I think the majority will use the standard response text.

carloshorn:
The proposed feature is already included in *responses*. You can define a custom reason using a `response_callback`.

Example:
```python
def custom_reason(response):
    if response.status_code == 299:
        response.reason = 'My Reason'
    return response

with responses.RequestsMock(response_callback=custom_reason) as reqmock:
    r = responses.Response(responses.GET, 'https://www.example.com', status=299)
    reqmock.add(r)
    print(requests.get('https://www.example.com').reason)
```

Issue #255, opened 2019-02-08
Add responses.__version__


zborboa-g:
Add `__version__` attribute to `responses` so that `responses.__version__` returns the module's version.

Example implementation is available in the requests library (https://github.com/requests/requests/).
- Create `responses/__version__.py` that contains `__version__ = ...`
- In `responses/__init__.py` import `__version__`: `from .__version__ import __version__`
- In `responses/setup.py` read `__version__.py` and use when calling `setup()`

https://github.com/requests/requests/blob/master/requests/__version__.py
https://github.com/requests/requests/blob/master/requests/__init__.py
https://github.com/requests/requests/blob/master/setup.py

beliaev-maksim:
@markstory 
I believe that this one could be closed according to comments in #332

PR #332, opened 2020-06-23
Added __title__, __description__, and __version__ to responses.

hramezani:
Fixes https://github.com/getsentry/responses/issues/255


dcramer:
What is the purpose of these?

You can get the version of libraries via `pkg_resources` already, and afaik the other attributes aren't used?

hramezani:
>What is the purpose of these?

Some libraries like `requests`, `django`, `flask`, ... has a `__version__` or `get_version()` or both to fetch the library version.
Also, it can be a starting point to improve the library structure and move `responses.py` to a directory and maybe split it to multiple files.(to make it more readable)

> You can get the version of libraries via `pkg_resources` already, and afaik the other attributes aren't used?
Yes, you are right. 
> afaik the other attributes aren't used?
Yes, we can remove them. TBH, I just added them because they are in `requests` library.

BTW, If this change does not make sense to you we can close the PR.

dcramer:
I've used ``__version__`` in the past, but its another point that has to be continuously updated and I feel like it has less values these days.

That said, I've never used any of the others in the hundreds of libraries I've released. I don't know that we need them, and I'd encourage to us to avoid breaking up the file unless it makes sense (because its gotten unwieldy). I certainly don't maintain things anymore, but the single file has the advantage of being easy to understand scope and see how everything works.

Just my two cents.

hramezani:
@dcramer Thanks for the explanation.
I removed the ` __title__` and `__description__`.
Again, If it does not make sense to you we can close the PR and also issue #255.

PR #334, opened 2020-06-26
Omit test files from coverage

Lewiscowles1986:
It actually looks like there is dead code in the tests if only 96% is the result.

In any case the coverage of the library should be what is reported.

Coverage of the library is artificially inflated from tests out of the box.

This fixes that

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,2 @@
+[run]
+omit = test*


Lewiscowles1986:
[OMD](https://codecov.io/gh/getsentry/responses/compare/647a0f3e0d634755503af21c2fe52421f13a9d4b...fc94a9178af08076d4a015091d353b3173d5fc91) 😂 yes the coverage will go down if it is reported accurately...

dcramer:
Coverage should include tests, because tests are code as well. If the tests aren't fully covered, it means they're not being executed. If you want to only look at coverage of the non-test code, you can do that, but thats unrelated to tracking the data.

Lewiscowles1986:
🙄

PR #336, opened 2020-07-03
Add matchers for post keys

### LINKED ISSUE #335: `assert_call_count` should be able to check urls without matching querystring
When adding an url in `responses`, querystring is not matched by default:

```python
@responses.activate
def test_my_test():
    responses.add(responses.GET, "http://www.example.com")
    # Both these requests will be captured
    requests.get("http://www.example.com")
    requests.get("http://www.example.com?hello=world")
```

`assert_call_count` doesn't reproduce this behaviour:
```python
@responses.activate
def test_assert_call_count_always_match_qs():
    responses.add(responses.GET, "http://www.example.com")
    requests.get("http://www.example.com")
    requests.get("http://www.example.com?hello=world")

    # One call on each url, querystring is matched by default
    responses.assert_call_count("http://www.example.com", 1) is True
```

I would like to be able to check the call count on one particular url without matching querystring, eg.:

```python
@responses.activate
def test_assert_call_count_match_querystring():
    responses.add(responses.GET, "http://www.example.com")
    requests.get("http://www.example.com")
    requests.get("http://www.example.com?hello=world")

    responses.assert_call_count("http://www.example.com", 2, match_querystring=False) is True
    responses.assert_call_count("http://www.example.com?hello=world", 2, match_querystring=False) is True
```

pmourlanne:
Closes #335 

I'm not really satisfied with the external API:
- `responses.assert_call_count(url, 2, match_querystring=False)` is a bit clunky
- `match_querystring` should maybe be `False` by default to match capture behaviour? But then the change is not backwards compatible :o

The internal change with `call_url_matches` is not really clean either, I don't know how to improve it :\

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -241,12 +241,31 @@ def urlencoded_params_matcher(params):
     return match
 
 
+def urlencoded_keys_matcher(keys):
+    def match(request_body):
+        return sorted(keys) == sorted(dict(parse_qsl(request_body)).keys())
+
+    return match
+
+
 def json_params_matcher(params):
     def match(request_body):
         try:
             if isinstance(request_body, bytes):
                 request_body = request_body.decode("utf-8")
             return params == json_module.loads(request_body)
+        except JSONDecodeError:
+            return False
+
+    return match
+
+
+def json_keys_matcher(keys):
+    def match(request_body):
+        try:
+            if isinstance(request_body, bytes):
+                request_body = request_body.decode("utf-8")
+            return sorted(keys) == sorted(json_module.loads(request_body).keys())
         except JSONDecodeError:
             return False
 
--- a/test_responses.py
+++ b/test_responses.py
@@ -1228,3 +1228,39 @@ def test_request_matches_post_params():
 
     run()
     assert_reset()
+
+
+def test_request_matches_post_keys():
+    @responses.activate
+    def run():
+        responses.add(
+            method=responses.POST,
+            url="http://example.com/",
+            body="one",
+            match=[responses.json_keys_matcher(["from", "to"])],
+        )
+        responses.add(
+            method=responses.POST,
+            url="http://example.com/",
+            body="two",
+            match=[responses.urlencoded_keys_matcher(["from", "to"])],
+        )
+
+        resp = requests.request(
+            "POST",
+            "http://example.com/",
+            headers={"Content-Type": "x-www-form-urlencoded"},
+            data={"from": 12, "to": 24},
+        )
+        assert_response(resp, "two")
+
+        resp = requests.request(
+            "POST",
+            "http://example.com/",
+            headers={"Content-Type": "application/json"},
+            json={"from": 12, "to": 24},
+        )
+        assert_response(resp, "one")
+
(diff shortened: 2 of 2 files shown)


markstory:
> match_querystring should maybe be False by default to match capture behaviour? But then the change is not backwards compatible :o

I think assertions should default to strict matching, and disabling querystring matching relaxes the assertion which is a good approach. :+1:

 > responses.assert_call_count(url, 2, match_querystring=False) is a bit clunky

We could have a querystring lax assertion as well instead of a querystring parameter?

responses.py
```diff
@@ -692,14 +692,18 @@ def stop(self, allow_assert=True):
...
-                for call in self.calls
-                if call.request.url == _ensure_url_default_path(url)
-            ]
-        )
+    def assert_call_count(self, url, count, match_querystring=True):
+        def call_url_matches(call):
+            call_url = call.request.url
+            if match_querystring:
+                return call_url == _ensure_url_default_path(url)
+            else:
```
markstory, line 700: You could skip this else as the if block returns :golf:

pmourlanne:
> We could have a querystring lax assertion as well instead of a querystring parameter?

Great idea :+1: What should we call it? `assert_call_count_no_qs`?

markstory:
I am wondering if we should use the idea of matches from #319 here as well instead of adding more methods. 🤔

pmourlanne:
> I am wondering if we should use the idea of matches from #319 here as well instead of adding more methods

In my original use case, an URL is called with qs parameters I don't particularly care about, and I don't know in advance the querystring *values*. I know the *keys* in the parameters.\
I guess we can extend the matches from #319 to match on keys only, ignoring values. Would that be OK with you?

pmourlanne:
Now that https://github.com/getsentry/responses/pull/319 is merged, I implemented new matchers only matching on URL / JSON param *keys*

It fits my use case and is lightweight :)

responses.py
```diff
@@ -253,6 +260,18 @@ def match(request_body):
     return match
 
 
+def json_keys_matcher(keys):
```
markstory, line 263: I wonder if all the matchers should be moved into another module. Perhaps `responses.match`? That could be done independently of these changes too though.

beliaev-maksim:
Hi @pmourlanne,

will current implementation sort your issue out?
https://github.com/getsentry/responses#matching-request-parameters

if yes, you can consider closing #336 and #335

pmourlanne:
I'll try and have a look soonish :)

pmourlanne:
@beliaev-maksim This PR is now redundant with what's in `main`, that's clear :+1: I'll close this PR.

Although the initial need from https://github.com/getsentry/responses/issues/335 is still present. (I don't know how / why my PR turned out this way, that's certainly my bad :>)  
I want to be able to capture on a URL regardless of the querystring (that's possible), and assert that the URL was called regardless of the querystring (that part's not possible)

beliaev-maksim:
@pmourlanne 
thanks for the feedback, then we may have a look at the issue later :)

PR #347, opened 2020-09-09
Allow matching on empty request body

phillipuniverse:
If there was no body sent in the original request it comes into the matcher as `None`.

### DIFF
--- a/test_responses.py
+++ b/test_responses.py
@@ -1240,8 +1240,22 @@ def test_request_matches_empty_body():
             match=[responses.json_params_matcher(None)],
         )
 
+        responses.add(
+            method=responses.POST,
+            url="http://example.com/",
+            body="two",
+            match=[responses.urlencoded_params_matcher(None)],
+        )
+
         resp = requests.request("POST", "http://example.com/",)
         assert_response(resp, "one")
+
+        resp = requests.request(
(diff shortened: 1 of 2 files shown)


phillipuniverse:
Took a few tries for a passing build but this should be good to review now!

hramezani:
Do we need to handle `None` in [urlencoded_params_matcher](https://github.com/getsentry/responses/blob/master/responses.py#L237) as well?

phillipuniverse:
Good question and probably, I'll add some tests around that.

phillipuniverse:
@hramezani good call there was an issue there too. Ready for re-review!

test_responses.py
```diff
@@ -1228,3 +1228,34 @@ def run():
...
+        )
+
+        responses.add(
+            method=responses.POST,
+            url="http://example.com/",
+            body="two",
+            match=[responses.urlencoded_params_matcher(None)],
+        )
+
+        resp = requests.request("POST", "http://example.com/",)
```
hramezani, line 1250:
```suggestion
        resp = requests.request("POST", "http://example.com/")
```

hramezani (approved):
LGTM.
@markstory Please take a look if you have time.

Issue #345, opened 2020-09-04
Error message could be more specific when POST request fails to match expected data parameters


dannguyen:
I believe this is related to the feature added in this pull request: https://github.com/getsentry/responses/pull/336

When a test attempts a POST request with different `data` values than mocked via `responses.add(match=[...])`, e.g.

```py

@responses.activate
def test_foobar():
    responses.add('POST',
       'http://example.com',
        body='hello',
        match=[
            responses.urlencoded_params_matcher({'id': 'good'})
        ]
    )

    requests.post('http://example.com', data={'id': 'bad'})
```

I get a `ConnectionError` that doesn't specify that the problem stems from the `matches` parameter – basically, the error message is less than helpful:

```
E           requests.exceptions.ConnectionError: Connection refused by Responses - the call doesn't match any registered mock.
E           
E           Request: 
E           - POST http://example.com/
E           
E           Available matches:
E           - POST http://example.com/

../../.pyenv/versions/3.8.5/lib/python3.8/site-packages/responses.py:689: ConnectionError
```

Not a big deal at all, just an inconvenience. But I was wondering if maybe I was just doing/expecting something wrong, rather than this being something that needs a fix/improvement.

PR #349, opened 2020-10-07
Improve ConnectionError error message.

hramezani:
Fixes https://github.com/getsentry/responses/issues/345

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -357,15 +357,15 @@ class BaseResponse(object):
 
     def matches(self, request):
         if request.method != self.method:
-            return False
+            return False, "Method does not match"
 
         if not self._url_matches(self.url, request.url, self.match_querystring):
-            return False
+            return False, "URL does not match"
 
         if not self._body_matches(self.match, request.body):
-            return False
+            return False, "Parameters do not match"
 
-        return True
+        return True, ""
 
 
 class Response(BaseResponse):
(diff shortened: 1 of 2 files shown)


responses.py
```diff
@@ -355,15 +355,15 @@ def get_response(self, request):
...
-            return False
+            return False, "Method does not match"
 
         if not self._url_matches(self.url, request.url, self.match_querystring):
-            return False
+            return False, "URL does not match"
 
         if not self._body_matches(self.match, request.body):
-            return False
+            return False, "Parameters does not match"
```
markstory, line 364:
```suggestion
            return False, "Parameters do not match"
```

test_responses.py
```diff
@@ -1259,3 +1259,25 @@ def run():
...
+            "http://example.com",
+            match=[responses.urlencoded_params_matcher({"foo": "bar"})],
+        )
+
+        with pytest.raises(ConnectionError) as excinfo:
+            requests.post("http://example.com", data={"id": "bad"})
+        msg = str(excinfo.value)
+        assert "- POST http://example1.com/ URL does not match" in msg
+        assert "- GET http://example.com/ Method does not match" in msg
+        assert "- POST http://example.com/ Parameters does not match" in msg
```
markstory, line 1280:
```suggestion
        assert "- POST http://example.com/ Parameters do not match" in msg
```

responses.py
hramezani: Thanks @markstory. Fixed!

test_responses.py
hramezani: Thanks @markstory. Fixed!

hramezani:
@markstory Thanks for the review. I addressed your comments.

PR #350, opened 2020-10-10
Allow request.params to have multiple values for the same key

tiqwab:
After incorporating https://github.com/getsentry/responses/pull/301, responses can return requested parameters by `responses.calls[].request.params`, but it doesn't support multiple values for the same parameter for now.

Although there is no standard about the multiple values for the same request parameter, requests can accept list as value of `params` (ref. [Passing Parameters In URLs](https://requests.readthedocs.io/en/master/user/quickstart/#passing-parameters-in-urls)), so I feel it is better to handle it in the same way.

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -5,6 +5,8 @@ import inspect
 import json as json_module
 import logging
 import re
+from itertools import groupby
+
 import six
 
 from collections import namedtuple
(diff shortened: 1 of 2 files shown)


markstory:
Thank you 👍

Issue #354, opened 2020-11-10
Import error with new version of urllib3


epwalsh:
Hi, I'm getting this error when using responses:

```
tests/common/file_utils_test.py:9: in <module>
    import responses
/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/responses.py:29: in <module>
    from urllib3.connection import HTTPHeaderDict
E   ImportError: cannot import name 'HTTPHeaderDict' from 'urllib3.connection' (/opt/hostedtoolcache/Python/3.8.6/x64/lib/python3.8/site-packages/urllib3/connection.py)
```

This is with responses 0.12 and urrlib3 1.26.0.

pbelskiy:
+1 

https://github.com/pbelskiy/helix-swarm/runs/1382148223

```
.tox/pytest/lib/python3.8/site-packages/responses.py:27: in <module>
69
    from requests.packages.urllib3.connection import HTTPHeaderDict
70
E   ImportError: cannot import name 'HTTPHeaderDict' from 'urllib3.connection' (/home/runner/work/helix-swarm/helix-swarm/.tox/pytest/lib/python3.8/site-packages/urllib3/connection.py)
```

hramezani:
I created a PR to fix this import error https://github.com/getsentry/responses/pull/355

PR #355, opened 2020-11-10
Fixed HTTPHeaderDict import error.

hramezani:
Fixes https://github.com/getsentry/responses/issues/354

### DIFF
--- a/responses.py
+++ b/responses.py
@@ -28,7 +28,7 @@ except ImportError:
 try:
     from requests.packages.urllib3.connection import HTTPHeaderDict
 except ImportError:
-    from urllib3.connection import HTTPHeaderDict
+    from urllib3.response import HTTPHeaderDict
 
 if six.PY2:
     from urlparse import urlparse, parse_qsl, urlsplit, urlunsplit


epwalsh:
This would probably break for older versions of urllib3 then, right? Might need to pin urllib3 to `>=1.26.0` in `setup.py`.

hramezani:
> This would probably break for older versions of urllib3 then, right? Might need to pin urllib3 to `>=1.26.0` in `setup.py`.

Thanks, @epwalsh, I pinned it to `>=1.26.0`

responses.py
```diff
@@ -28,7 +28,7 @@
 try:
     from requests.packages.urllib3.connection import HTTPHeaderDict
 except ImportError:
-    from urllib3.connection import HTTPHeaderDict
+    from urllib3._collections import HTTPHeaderDict
```
markstory, line 31: I don't know about this. My understanding is that modules with `_` prefixes are internal and shouldn't be used by consumers of a package. I'm going to open an issue in urllib3 about the removed import causing a compatibility break.

hramezani: I changed the patch to import `HTTPHeaderDict` from  `urllib3.response` based on https://github.com/urllib3/urllib3/issues/2045#issuecomment-725497199

pquentin (approved):
Thanks for this, @hramezani! Fetching from `urllib3.response` is going to work on all versions of urllib3 that actually have `HTTPHeaderDict`

PR #360, opened 2020-11-14
Improve `replace` error message.

hramezani:


test_responses.py
```diff
@@ -132,14 +132,31 @@ def run():
...
-            responses.replace(responses.GET, replacement)
+        if isinstance(original, BaseResponse):
+            responses.add(original)
+        else:
+            responses.add(responses.GET, original)
+
+        with pytest.raises(ValueError) as excinfo:
+            if isinstance(replacement, BaseResponse):
+                responses.replace(replacement)
+            else:
```
markstory, line 152: This test is getting complicated. It might be simpler to split out these object based tests into a separate test.

hramezani: I created a separate test function to test replace error on `Response` object.

hramezani:
@markstory Could you please take a look at my fix?
I am going to work on another issue and this one is needed.
Thanks

PR #362, opened 2020-12-04
Documentation: Fix an examples in README.rst

SergiyKolesnikov:
Fix the `test_calc_api()` example in "Matching Request Parameters"
    
- `body` paramater of `responses.add()` does not accept arguments of type
  `int`. Changed to `str`.
- Values in the request parameters given to
  `responses.urlencoded_params_matcher()` must be of type `str`. Otherwise,
  `ConnectionError` is thrown, because of not matching parameters.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -148,9 +148,9 @@ other formats.
         responses.add(
             responses.POST,
             url='http://calc.com/sum',
-            body=4,
+            body="4",
             match=[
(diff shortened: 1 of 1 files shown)


hramezani (approved):
Thanks :+1:

markstory:
Thank you 👍

PR #363, opened 2020-12-15
Add upsert method.(#227)

### LINKED ISSUE #227: Add an "update" (upsert) method for adding responses to RequestMock
RequestMock currently has add, remove and replace.

However replace doesn't work if the result does not already exist, meaning you need to write a lot of remove/add code (which does work if the result does not already exist) when updating a RequestMock.

Ideally there would be an "upsert" type method which would wrap up this type of operation, leading to tighter test code.

Due to the ordering of RequestMock, a directionality would probably be needed for common use cases - i.e. add to the end if does not exist, or add to the beginning if does not exist.

hramezani:
Fixed #227

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -475,6 +475,10 @@ replaced.
         assert resp.json() == {'data': 2}
 
 
+The ``upsert`` function allows a previously registered ``response`` to be
+changed like ``replace``. If the response is registered, the ``upsert`` function
+will registered it like ``add``.
+
 ``remove`` takes a ``method`` and ``url`` argument and will remove **all**
 matched responses from the registered list.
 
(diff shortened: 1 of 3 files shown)


responses.py
```diff
@@ -621,6 +621,20 @@ def replace(self, method_or_response=None, url=None, body="", *args, **kwargs):
             raise ValueError("Response is not registered for URL %s" % url)
         self._matches[index] = response
 
+    def upsert(self, method_or_response=None, url=None, body="", *args, **kwargs):
+        """
+        Replaces a response previously added using ``add()`` or adds it if it does
+        not exists. The signature is identical to ``add()``. The response is identified
+        using ``method``and ``url``, and the first matching response is replaced.
```
markstory, line 628:
```suggestion
        Replaces a response previously added using ``add()``, or adds the response
        it if no response exists.  Responses are matched using ``method``and ``url``.
        The first matching response is replaced.
```

hramezani: Done!

PR #365, opened 2020-12-27
Add type stubs

### LINKED ISSUE #339: Adding type hints
I was wondering if you'd be open to adding types directly to the repo.

If not would you be okay with me adding them to the [typeshed](https://github.com/python/typeshed/)?

sbdchd:
Add type stubs for the library as a separate stub file which means we can
use the python 3 syntax, with the downside of having to keep them in
sync with the source.

I ran [monkeytype][1] and [stubgen][2] over responses.py and combined the
results manually.

I had to move the module into a package for the types to work.
see: https://www.python.org/dev/peps/pep-0561/#packaging-type-information

Fixes #339

https://www.python.org/dev/peps/pep-0561/

[1]: https://github.com/Instagram/MonkeyType
[2]: https://mypy.readthedocs.io/en/stable/stubgen.html

### DIFF
--- a/responses/__init__.pyi
+++ b/responses/__init__.pyi
@@ -1,0 +1,285 @@
+from collections import Sequence, Sized
+from typing import (
+    Any,
+    Callable,
+    Iterator,
+    Mapping,
+    Optional,
+    Pattern,
+    NamedTuple,
+    Protocol,
+    TypeVar,
+)
+from io import BufferedReader, BytesIO
+from re import Pattern
+from requests.adapters import HTTPResponse, PreparedRequest
+from requests.cookies import RequestsCookieJar
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing_extensions import Literal
+from unittest import mock as std_mock
+from urllib.parse import quote as quote
+from urllib3.response import HTTPHeaderDict
+
+JSONDecodeError = ValueError
+
+def _clean_unicode(url: str) -> str: ...
+def _cookies_from_headers(headers: Dict[str, str]) -> RequestsCookieJar: ...
+def _ensure_str(s: str) -> str: ...
+def _ensure_url_default_path(
+    url: Union[Pattern[str], str]
+) -> Union[Pattern[str], str]: ...
+def _handle_body(
+    body: Optional[Union[bytes, BufferedReader, str]]
+) -> Union[BufferedReader, BytesIO]: ...
+def _has_unicode(s: str) -> bool: ...
+def _is_string(s: Union[Pattern[str], str]) -> bool: ...
+def get_wrapped(
+    func: Callable[..., Any], responses: RequestsMock
+) -> Callable[..., Any]: ...
+def json_params_matcher(
+    params: Optional[Dict[str, Dict[str, str]]]
+) -> Callable[..., Any]: ...
+def urlencoded_params_matcher(
+    params: Optional[Dict[str, str]]
+) -> Callable[..., Any]: ...
+
+class Call(NamedTuple):
+    request: PreparedRequest
+    response: Any
+
+_Body = Union[str, BaseException, "Response", BufferedReader, bytes]
+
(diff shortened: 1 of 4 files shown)


responses/__init__.pyi
```diff
@@ -0,0 +1,285 @@
...
+        self,
+        method: str,
+        url: Union[Pattern[str], str],
+        callback: Callable[[Any], Any],
+        stream: bool = ...,
+        content_type: Optional[str] = ...,
+        match_querystring: bool = ...,
+        match: List[Any] = ...,
+    ) -> None: ...
+    def get_response(  # type: ignore [override]
```
sbdchd, line 129: I used the following config when developing, not sure if you want mypy running or not in CI


```ini
[mypy]
show_column_numbers=True
show_error_codes = True

disallow_any_unimported=True
disallow_any_expr=False
disallow_any_decorated=True
disallow_any_explicit=False
disallow_any_generics=True
disallow_subclassing_any=True

disallow_untyped_calls=True
disallow_untyped_defs=True
disallow_incomplete_defs=True
check_untyped_defs=True
disallow_untyped_decorators=True

no_implicit_optional=True
strict_optional=True

warn_redundant_casts=True
warn_unused_ignores=True
warn_no_return=True
warn_return_any=False
warn_unreachable=False

strict_equality=True

ignore_missing_imports=False

[mypy-urllib3.*]
ignore_missing_imports=True

[mypy-test_responses]
disallow_untyped_defs=False
disallow_untyped_calls=False
disallow_any_decorated=False
```

responses/__init__.pyi
```diff
@@ -0,0 +1,285 @@
...
+    def start(self) -> None: ...
+    def stop(self, allow_assert: bool = ...) -> None: ...
+    def assert_call_count(self, url: str, count: int) -> bool: ...
+
+_F = TypeVar("_F", bound=Callable[..., Any])
+
+class _Activate(Protocol):
+    def __call__(self, func: _F) -> _F: ...
+
+class _Add(Protocol):
```
sbdchd, line 184: I ended up doing this really ugly protocol stuff since there isn't really a good way to get the function type out of the class  for reusing below.

Once python 2 support is removed, add the types are added inline, we won't need any of this Protocol stuff

setup.py
```diff
@@ -62,14 +62,15 @@ def run_tests(self):
     license="Apache 2.0",
     long_description=open("README.rst").read(),
     long_description_content_type="text/x-rst",
-    py_modules=["responses"],
+    packages=["responses"],
```
sbdchd, line 65: required by https://www.python.org/dev/peps/pep-0561/#packaging-type-information

responses/__init__.pyi
```diff
@@ -0,0 +1,285 @@
...
+        url: Optional[Union[Pattern[str], str]] = ...,
+        body: _Body = ...,
+        json: Optional[Any] = ...,
+        status: int = ...,
+        headers: Optional[Mapping[str, str]] = ...,
+        stream: bool = ...,
+        content_type: Optional[str] = ...,
+        adding_headers: Optional[Mapping[str, str]] = ...,
+        match_querystring: bool = ...,
+        match: List[Any] = ...,
```
sbdchd, line 197: I changed the methods that take `*args` and `**kwargs` to list out the specific parameters that are used instead to help with typos

responses/__init__.pyi
markstory: Having mypy run as part of CI would be ideal. I can take care of that though.

markstory: How were you running mypy locally? When I ran it with this configuration file there were ~380 errors from the test_responses module.

sbdchd: not sure what command I was running, but were you using the config above?

```
[mypy-test_responses]
disallow_untyped_defs=False
disallow_untyped_calls=False
disallow_any_decorated=False
```

should ignore most of the errors until the tests module is fully typed

PR #369, opened 2021-02-19
Add an exemple of unittest usage

romuald:
This is a simple example of usage inside a classic unittest setUp() clause

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -331,6 +331,31 @@ Responses as a pytest fixture
         resp = requests.get('http://twitter.com/api/1/foobar')
         assert resp.status_code == 200
 
+Responses inside a unittest setUp()
+-----------------------------------
+
+When run with unittest tests, this can be used to set up some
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -331,6 +331,28 @@ Responses as a pytest fixture
...
         assert resp.status_code == 200
 
+Responses inside a unittest setUp()
+-----------------------------------
+
+.. code-block:: python
+
+    def setUp():
+        self.responses = responses.RequestsMock()
+        self.responses.start()
```
markstory, line 341: It would be good to give some context on this example and why you would use this instead of using the decorators. Perhaps explaining that this approach can be used when multiple test methods need to use the same mocks?

romuald: Good idea. I've tried to add a description of the example, but I'm not sure the text is correct

(In the meanwhile I've switched to using pytest fixtures ^^)

markstory:
Thank you 👍

PR #370, opened 2021-02-23
Improve Responses representation for debug purposes

### LINKED ISSUE #368: Facilitate how to view the registered responses for debugging purposes
Hi maintainers! This is my first issue and intent to help here 😄 

## Motivation
When writing my tests and debugging them, sometimes I want to know which are the registered responses so far since we tend to use different fixtures to add responses and we lose track of which are the URLs expected to be called before the test ends.

The documentation says that in order to get some access to them we should access them via `responses.mock._matches` which in turn, it will return the list of `Response` objects and we have to access them to have information such as the precious URL, as in:
```
[
    <responses.Response object at 0x10a121670>,
    <responses.Response object at 0x10a1216a0>,
    <responses.Response object at 0x10a1216d0>
]
```

## Suggestion
To have a method that facilitates retrieving information of which are these registered responses.

I'd suggest something like `responses.mock.show_registered(urls_only: bool = False)` and return a list of the URLs registered if `urls_only == True` or a dictionary structure nested by `method` and `url`.
For example having:
```
responses.add(
    responses.GET, "http://example.com/", body="", headers={"X-Test": "foo"}
)
responses.add(
    responses.GET, "http://example.com/", body="", headers={"X-Test": "foo"}
)
responses.add(
    responses.POST, "http://anotherdomain.com/", body={"field": "value"}
)
```
then calling `responses.mock.show_registered(urls_only=True)` would return:
```
["http://example.com/", "http://example.com/", "http://anotherdomain.com/" ]
```
or by calling just `responses.mock.show_registered()` would return:
```
{
    'GET': {
        'http://example.com/': [
            {
                'body': '',
                'content_type': 'text/plain',
                'headers': {'X-Test': 'foo'},
                'status': 200
            },
            {
                'body': '',
                'content_type': 'text/plain',
                'headers': {'X-Test': 'foo'},
                'status': 200
            }
        ]
    },
    'POST': {
        'http://anotherdomain.com/': [
            {
                'body': {'field': 'value'},
                'content_type': 'text/plain',
                'headers': None,
                'status': 200
            }
        ]
    }
}
```

## Implementation
I have already implemented this in a local branch, if you think it could be useful I can create a Pull Request for you to review it and give me some feedback.

mromagnoli:
Closes #368 

----

In order to help devs while debugging the list of mocked Responses:
- I added an implementation for `Response`'s `__str__` and `__repr__` so is clearer which are the attributes for that Response object
- Also added a "proxy" method to `responses.mock._matches` in `RequestMock` to list the already registered mocked Responses by calling `responses.mock.registered()` 

For example:
```
> response = Response(method=responses.GET, url="http://example.com/")
> response
<responses.Response: http://example.com/>
> print(response)
{"url": "http://example.com/", "content_type": "text/plain", "headers": null, "status": 200}
```

**Docs**: I was thinking of updating the docs once maintainers think the solution is OK


mromagnoli:
I have to leave for today, I'll check tomorrow all the failing checks 👍

mromagnoli:
In order to avoid the mypy error that is happening in all other PRs and master, we have to add `upsert` to the list of stubs in `__init__.pyi`

responses/__init__.py
```diff
@@ -418,6 +418,22 @@ def get_response(self, request):
...
+                    "content_type": self.content_type,
+                    "headers": self.headers,
+                    "status": self.status,
+                    "url": self.url,
+                },
+                sort_keys=True,
+            )
+        )
+
+    def __repr__(self):
```
markstory, line 434: repr is typically intended to output something that could be put back into a repl to create the object once again. I think we either shouldn't implement it or implement it in a way that it could be evaled again. This [article](https://www.journaldev.com/22460/python-str-repr-functions#what8217s-the-difference-between--str-and--repr) has an example of how repr and str could differ.

responses/__init__.py
```diff
@@ -418,6 +418,22 @@ def get_response(self, request):
             preload_content=False,
         )
 
+    def __str__(self):
+        return str(
+            json_module.dumps(
```
markstory, line 423: Could we include a prefix so that it is clearer that the object is not actually a dict/json? Something like `f"Response(url={self.url} status={self.status} content_type={self.content_type} headers={json.dumps(self.headers)})"` could work and might be a way to approach repr too.

responses/__init__.py
mromagnoli: Totally agree, in this case, I went for the second option mentioned on the [docs](https://docs.python.org/3/reference/datamodel.html#object.__repr__):

_If at all possible, this should look like a valid Python expression that could be used to recreate an object with the same value (given an appropriate environment). If this is not possible, a string of the form <...some useful description...> should be returned._

I believe that the Response's URL is, for most of the cases, the most used attribute and it could be easily checked when we call, for instance, `Responses.mock.registered()` or just displaying a Response instance in a debugging session, as in:
```
> response = Response(method=responses.GET, url="http://example.com/")
> response
<responses.Response: http://example.com/>
```

What do you think?

responses/__init__.py
mromagnoli: I like this approach, perhaps we can implement `__repr__` like this and not implement `__str__`, we just leave the same result for both. I think this could work

mromagnoli: I was wondering if, perhaps, it is a good idea to have this representation for `__str__` and another one similar but including `body` for `__repr__`, like this we would have everything to recreate the same object as is expected for `__repr__`. `body` could be large to add it always for `__str__`

responses/__init__.py
markstory: I agree that the URL is the most important attribute followed by the HTTP method.

mromagnoli:
To update, there `__repr__` and `__str__` are returning the same since just `__repr__` is implemented.
The format is:
```
"<Response(url='http://example.com/' status=200 content_type='text/plain' headers='null')>"
```
and `responses.mock.registered()` lists responses as:
```
ipdb> responses.mock.registered()

[<Response(url='http://example.com/' status=404 content_type='text/plain' headers='{"X-Test": "foo"}')>,
<Response(url='http://example.com/' status=200 content_type='text/plain' headers='{"X-Test": "foo"}')>, 
<Response(url='http://anotherdomain.com/' status=200 content_type='text/plain' headers='null')>]
```

mromagnoli:
Hi @markstory  👋  how do you feel about this last version? Do you think we are good to go?

responses/__init__.py
```diff
@@ -804,6 +818,7 @@ def assert_call_count(self, url, count):
     "DELETE",
     "GET",
     "HEAD",
+    "list",
```
markstory, line 821: Should this be `registered`?

responses/test_responses.py
```diff
@@ -1397,3 +1397,78 @@ def run():
...
+        )
+        third_response = Response(
+            responses.POST,
+            "http://anotherdomain.com/",
+        )
+        responses.add(first_response)
+        responses.add(second_response)
+        responses.add(third_response)
+
+        mocks_list = responses.mock.registered()
```
markstory, line 1468: Could this also be `responses.registered()` based on the export aliases?

responses/__init__.py
mromagnoli: you're right 👍 , I forgot to rename it

responses/test_responses.py
mromagnoli: totally true, I'll change it. Thanks

markstory:
Thank you 👍

PR #372, opened 2021-02-26
Fix mypy lint failure by adding 'upsert' to mypy stubs

mromagnoli:
Master branch and the later PRs were failing with `mypy` error 
```
Module has no attribute "upsert"  [attr-defined]
```
I added the stub in the `__init__.pyi` file to solve it and get our checks green again.


markstory:
Thank you 👍

PR #375, opened 2021-03-04
Allow calls to requsts between calls to responses.add

Jlrine2:
This change allows the use of multiple responses intermixed by calls to requests by removing previous matches that have already been called if multiple exist.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -669,7 +669,7 @@ class RequestsMock(object):
         return get_wrapped(func, self)
 
     def calls_for_match(self, match):
-        return len([call for call in self.calls if call.request.url == match.url])
+        return len([call for call in self.calls if match.matches(call.request)])
 
     def _find_match(self, request):
         found = None
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -668,7 +668,7 @@ class RequestsMock(object):
     def activate(self, func):
         return get_wrapped(func, self)
 
-    def calls_for_match(self, match):
+    def _calls_for_match(self, match):
         return len([call for call in self.calls if match.matches(call.request)])
 
     def _find_match(self, request):
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -682,7 +682,7 @@ class RequestsMock(object):
                     found = i
                     found_match = match
                 else:
-                    if self.calls_for_match(found_match) > 0:
+                    if self._calls_for_match(found_match) > 0:
                         # Multiple matches found, remove previously found matches
                         self._matches.remove(found_match)
                         found = i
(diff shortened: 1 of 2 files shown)


responses/__init__.py
```diff
@@ -668,6 +668,9 @@ def __exit__(self, type, value, traceback):
     def activate(self, func):
         return get_wrapped(func, self)
 
+    def calls_for_match(self, match):
+        return len([call for call in self.calls if call.request.url == match.url])
```
markstory, line 672: Should this also take into account the other kinds of matchers that responses has? For example encoded body and querystring matching?

Jlrine2: I'd suspect no but it could

Jlrine2: fixed @markstory

Jlrine2:
@markstory Is this good to go?

responses/__init__.py
```diff
@@ -679,8 +682,13 @@ def _find_match(self, request):
                     found = i
                     found_match = match
                 else:
-                    # Multiple matches found.  Remove & return the first match.
-                    return self._matches.pop(found), match_failed_reasons
+                    if self.calls_for_match(found_match) > 0:
```
markstory, line 685:
```suggestion
                    if self._calls_for_match(found_match) > 0:
```

responses/__init__.py
```diff
@@ -668,6 +668,9 @@ def __exit__(self, type, value, traceback):
     def activate(self, func):
         return get_wrapped(func, self)
 
+    def calls_for_match(self, match):
```
markstory, line 671:
```suggestion
    def _calls_for_match(self, match):
```

I'd rather we not have this as a public method.

responses/test_responses.py
```diff
@@ -1099,6 +1099,29 @@ def run():
...
+
+        responses.add(responses.GET, "http://example.com", body="best")
+        resp = requests.get("http://example.com")
+        assert_response(resp, "best")
+        # After all responses are used, last response should be repeated
+        resp = requests.get("http://example.com")
+        assert_response(resp, "best")
+
+    run()
+    assert_reset()
```
markstory, line 1122: I think we might be missing some coverage on multiple responses. These changes cause the following test to fail, while it current passes on master.

```python
def test_multiple_responses_out_of_order_two():
    @responses.activate
    def run():
        responses.add(responses.GET, "http://example.com", body="test")
        responses.add(responses.GET, "http://example.com", body="rest")
        responses.add(responses.GET, "http://example.com", body="fest")
        responses.add(responses.GET, "http://example.com", body="best")

        assert_response(resp, "test")

        assert_response(resp, "rest")

        assert_response(resp, "fest")

        assert_response(resp, "best")

        # After all responses are used, last response should be repeated
        assert_response(resp, "best")

    run()
    assert_reset()
```

beliaev-maksim:
@markstory 
we can close current PR as #455 is merged

PR #381, opened 2021-03-29
Fix erroneous typing for callback function

### LINKED ISSUE #379: Typing error: callback has incompatible type / ValueError
### Environment

Python 3.9.1
Reponses 0.13.1

### Steps to Reproduce

1. Updated responses from 0.13.0 to 0.13.1
2. mypy 0.812 typing check started to fail for following code:
```
    responses.add_callback(
        responses.POST,
        "…",
        callback=callback_function,
        content_type="application/json",
    )

def callback_function(
    request: PreparedRequest,
) -> tuple[int, dict[str, str], str]:
    …
        return (200, {}, json.dumps(some_structure))
```
with
```
error: Argument "callback" to "__call__" of "_AddCallback" has incompatible type "Callable[[PreparedRequest, VarArg(Any)], Tuple[int, Dict[str, str], str]]"; expected "Callable[..., str]"
```
3. After changing callback and its signature to string only, typing check passes, but test started to fail with following trace:
```
../../.pyenv/versions/registration-api/lib/python3.9/site-packages/responses/__init__.py:748: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <responses.CallbackResponse object at 0x107113700>
request = <PreparedRequest [POST]>

    def get_response(self, request):
        headers = self.get_headers()
    
        result = self.callback(request)
        if isinstance(result, Exception):
            raise result
    
>       status, r_headers, body = result
E       ValueError: too many values to unpack (expected 3)
```

### Expected Result

I would expect either typing check to pass or responses to work.

Thank you for your work on the responses project.

tobinus:
Since the type stub is not covered by the automated tests, I decided to change as little as possible. The new type hints for the callback should match with [how the callback is invoked and used](https://github.com/getsentry/responses/blob/master/responses/__init__.py#L445), though I have only tested it on a project I'm working on that makes use of the responses library.

EDIT: This should fix #379

### DIFF
--- a/responses/__init__.pyi
+++ b/responses/__init__.pyi
@@ -203,7 +203,7 @@ class _AddCallback(Protocol):
         self,
         method: str,
         url: Union[Pattern[str], str],
-        callback: Callable[..., str],
+        callback: Callable[[PreparedRequest], Union[Exception, Tuple[int, Mapping[str, str], _Body]]],
         match_querystring: bool = ...,
         content_type: Optional[str] = ...,
     ) -> None: ...


markstory:
Thank you 👍

PR #383, opened 2021-04-26
Wire into publish repo

chadwhitacre:
This enables `responses` to be released using the centralized [publish](https://github.com/getsentry/publish) repo.

Copy/pasta from `pytest-responses`.


chadwhitacre:
@markstory May also want [a post-deploy script](https://github.com/getsentry/craft#post-release-script-conventions) to version bump to dev (tbh I think that's frequent enough it should be baked into Craft but 🤷‍♂️).

markstory (approved):
Looks good to me. Just so I'm clear would releases be created with github actions going forward? I can clean up the travis based release builds too.

Issue #388, opened 2021-06-22
Passthroughs should override registered matches


tinvaan:
Overriding passthru essentially enables a user to add an exception to a generic rule (such as when using regexps). This especially useful when using callbacks. Consider the below example, for instance

```python
def callback_method(request):
    if request.path_url == "/allow/this/route":
        responses.add_passthru(request.url)
        r = requests.get(request.url)
        return (r.status_code, dict(r.headers), r.content)
    return (200, {"Content-Type": "application/json", json.dumps({"data": "mocked data"})})

responses.add_callback(
    responses.GET,
    re.compile("https://example.com"),
    callback=callback_method,
    content_type="application/json"
)

requests.get("https://example.com/mocked/route")   # Should return a mocked response
requests.get("https://example.com/allow/this/route") # Should return an actual response
```
Any thoughts? I have a fix tested and ready and would like to push it upstream.

carloshorn:
@tinvaan, I think #400 introduced a feature that may enable you to deal with your usecase.
Instead of adding new passthroughs in a callback during the test, you may just register a `PassthroughResponse`, for the specific requests at the beginning of the test. 
In my opinion, this approach would be cleaner, because you would not modify the state of your `RequestsMock` instance during the execution of your test, but instead simply provide responses for any expected requests before running your actual test lines.

PR #389, opened 2021-06-22
Make passthru's override registered matches

tinvaan:
Addresses https://github.com/getsentry/responses/issues/388

### DIFF
--- a/responses/test_responses.py
+++ b/responses/test_responses.py
@@ -1141,7 +1141,7 @@ def test_passthru(httpserver):
         resp = requests.get("http://example.com/two")
         assert_response(resp, "two")
         resp = requests.get("{}/one".format(httpserver.url))
-        assert_response(resp, "one")
+        assert_response(resp, "OK")
         resp = requests.get(httpserver.url)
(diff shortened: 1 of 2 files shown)


responses/test_responses.py
```diff
@@ -1141,7 +1141,7 @@ def run():
         resp = requests.get("http://example.com/two")
         assert_response(resp, "two")
         resp = requests.get("{}/one".format(httpserver.url))
-        assert_response(resp, "one")
+        assert_response(resp, "OK")
```
markstory, line 1144: This changing seems like a breaking change. This specific URL has a mock response that is now being ignored.

tinvaan: That's the intention here, right? The test asserts that the response being served is the live response and not the mock response.

This shows that the passthru overrides the registered mock.

markstory: Sure, but this will break behavior for existing users of responses which isn't something I'm willing to do at this point in time.

beliaev-maksim:
@markstory 
I think this PR could be closed, since #400 solves the issue

PR #390, opened 2021-07-24
rfc compliant url logic

dmitrypolo:
This addresses the fact that currently URLs are not matched using [RFCC-3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2) compliant host-names. When adding a mock URL such as `http://some-URL-HERE`, then we should expect that both `http://some-URL-HERE` and `http://some-url-here` should pass when submitting requests. Per the RFC:
> The host sub-component is case-insensitive.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -330,18 +330,14 @@ class BaseResponse(object):
 
             if match_querystring:
                 rfc_compliant_url = parse_url(url).url
-                rfc_result = self._url_matches_strict(rfc_compliant_url, other)
-                non_rfc_result = self._url_matches_strict(url, other)
-                return non_rfc_result or rfc_result
+                return self._url_matches_strict(rfc_compliant_url, other)
 
             else:
                 url_without_qs = url.split("?", 1)[0]
                 other_without_qs = other.split("?", 1)[0]
                 rfc_compliant_url_without_qs = parse_url(url_without_qs).url
 
-                rfc_result = rfc_compliant_url_without_qs == other_without_qs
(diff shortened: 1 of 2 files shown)


responses/__init__.py
```diff
@@ -323,13 +327,21 @@ def _url_matches(self, url, other, match_querystring=False):
                 url = _clean_unicode(url)
                 if not isinstance(other, six.text_type):
                     other = other.encode("ascii").decode("utf8")
+
             if match_querystring:
-                return self._url_matches_strict(url, other)
+                rfc_compliant_url = parse_url(url).url
+                rfc_result = self._url_matches_strict(rfc_compliant_url, other)
+                non_rfc_result = self._url_matches_strict(url, other)
```
markstory, line 334: Are there scenarios where the RFC compliant matching fails but the existing behavior does match?

dmitrypolo: Not that I can think of however, it should not make a difference. The final returned result is the `or` between the RFC match and the existing behavior. If say the RFC match failed but the existing behavior did succeed, then it would return `True` regardless.

markstory: Do we need the non-rfc compliant host comparisons then? Seems like the RFC comparisons are more relaxed than what we had before.

dmitrypolo: If this behavior is acceptable than no, I don't think the non-RFC comparison is needed.

markstory: I think it is fine. The case insensitive comparisons are how HTTP requests would work in the wild.

markstory:
Thank you 👍

PR #391, opened 2021-07-25
Fix matches types

Bachmann1234:
This method just takes in an optional dict. The implementation parses the json and compares the result. If we wanted to get really specific we could put in something more specific but the current type is just incorrect.

I'm very open to being wrong here but this is what im seeing in my testing


markstory:
Thank you 👍

Issue #394, opened 2021-08-06
AttributeError when urllib3 throws an IncompleteRead


FlMart:
**TL;DR:** There is a flag in `urllib3.HTTPResponse` that checks if the "Content-Length" header of a HTTP request matches its body size. If this flag is set, the error raised by a mismatch between the two causes an AttributeError in responses (see further below for details).
With the current version of requests, the required flag in urllib3 has to be patched in but it is also planned to be included in the requests library with their planned release of 3.0.0 (https://github.com/psf/requests/issues/4956#issue-405545488).

### Environment

python 3.8.5
requests==2.26.0
responses==0.13.3
urllib3==1.26.6

### Steps to Reproduce

Minimal working example:
```python
import unittest

import requests
import responses
import urllib3
from requests.exceptions import ChunkedEncodingError


def patch_urllib3():
    """Set urllib3's enforce_content_length to True by default."""
    previous_init = urllib3.HTTPResponse.__init__
    def new_init(self, *args, **kwargs):
        previous_init(self, *args, enforce_content_length = True, **kwargs)
    urllib3.HTTPResponse.__init__ = new_init

class TestResponses(unittest.TestCase):
    """Test that responses handles IncompleteRead exceptions correctly."""

    @responses.activate
    def test_raises_an_error(self):
        """Test that an IncorrectEncoding is thrown."""
        patch_urllib3()
        responses.add(**{
            "method"            : responses.GET,
            "url"               : "http://example.com/api/123",
            "body"              : "{'message': 'this body is too large'}",
            "adding_headers"    : {"content-length": "2"}
        })
        with self.assertRaises(ChunkedEncodingError):
            requests.get("http://example.com/api/123")
```

### Expected Result

When building a working example with fastapi and uvicorn, a `requests.exceptions.ChunkedEncodingError` is thrown. I would expect this to also happen when responses is used. If a simple
```python
def close(self):
    return
```
is added to `responses/__init__.py:OriginalResponseShim`, this is also the result that occurs with responses. 

### Actual Result

The actual result is an `AttributeError`:

```
Traceback (most recent call last):
  File "/local/home/fma/.conda/envs/endpoint/lib/python3.8/site-packages/urllib3/response.py", line 438, in _error_catcher
    yield
  File "/local/home/fma/.conda/envs/endpoint/lib/python3.8/site-packages/urllib3/response.py", line 541, in read
    raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
[...]

PR #396, opened 2021-08-18
fixes ChunkEnodingError when content length mismatches with the body content. 

ViAsmit:
Fixed Issue #394

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -13,7 +13,7 @@ from collections import namedtuple
 from functools import update_wrapper
 from requests.adapters import HTTPAdapter
 from requests.exceptions import ConnectionError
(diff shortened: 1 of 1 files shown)


responses/__init__.py
```diff
@@ -503,6 +503,11 @@ def __init__(self, headers):
     def isclosed(self):
         return True
 
+    def close(self):
+        raise ChunkedEncodingError(
+            "Connection broken: Incomplete Read, Content-length mismatch with body"
+        )
```
markstory, line 509: Won't raising an error here cause all requests to fail? The issue proposed having close() be a no-op, but without test coverage on this it is hard to tell what is working/not working.

beliaev-maksim:
@markstory 
please close this one

PR #397, opened 2021-08-26
provide request params as response argument

### LINKED ISSUE #371: Using urlencoded_params_matcher for matching query parameters
I'm getting started with `responses` (version `0.12.1`), and I'm trying to match query parameters. From the README, my understanding was that I could use `urlencoded_params_matcher` for that, but it seems that this (and the `match` argument of `add`) only works for POST data, but not for query args.

This is probably intended, but the documentation could be a bit clearer here.

It seems that the right way of doing what I want is to specify the query args in the URL. This is cumbersome due to the encoding, and if you mock a lot of URLs with slightly different query arg values then it gets confusing quickly. I would prefer to pass a dict instead whose content is automatically encoded (e.g. via `urllib.parse.urlencode`). Something like

```python
responses.add(
    method='GET',
    url='https://some-host/some-endpoint',
    params={
        'foo': 'bar',
        'x': 'y',
    },
)
```

which would be roughly equivalent to

```python
url = https://some-host/some-endpoint',
params = {
    'foo': 'bar',
    'x': 'y',
}
full_url = url + '?' + urlencode(params)  # construction should be more robust
responses.add(
    method='GET',
    url=full_url,
)
```

beliaev-maksim:
it will address #371 and similar requests

will be glad to hear feedback from you. Hope my change can slightly improve usability of this nice lib

Closes #371


responses/__init__.py
```diff
@@ -240,33 +240,6 @@ def _handle_body(body):
...
-        return (
-            params is None
-            if request_body is None
-            else sorted(params.items()) == sorted(parse_qsl(request_body))
-        )
-
-    return match
-
-
-def json_params_matcher(params):
```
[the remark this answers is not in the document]
beliaev-maksim: @markstory 
sure, I added imports and updated test to catch both implementations (current+deprecated)

beliaev-maksim:
Also,
Closes #107

responses/test_responses.py
```diff
@@ -891,13 +893,17 @@ def test_response_filebody():
...
         with responses.RequestsMock() as m:
-            with open("README.rst", "r") as out:
+            with open(current_file, "r") as out:
                 m.add(responses.GET, "http://example.com", body=out.read(), stream=True)
                 resp = requests.get("http://example.com")
-            with open("README.rst", "r") as out:
+            with open(current_file, "r") as out:
                 assert resp.text == out.read()
 
+    run()
```
beliaev-maksim, line 904: @markstory 
for some reason in this test was no call to the run()
I added call to enable this test, if it was intentional, then let me know and I will remove it back

beliaev-maksim: @markstory 
can you please trigger test run?

markstory:
Looking good, there are a few test failures that will need to be addressed before this can be merged.

beliaev-maksim:
@markstory 
now I see why this test was disabled, Python2 compiles data differently, so body will not match file content

I used six and enabled this test only for PY3

markstory:
Thank you 👍

beliaev-maksim:
@markstory 
great, thanks!

PR #398, opened 2021-08-29
Add flag to allow automatic calculation of the Content-Length header

djmattyg007:
To be correct, the Content-Length header must be calculated from the
binary representation of the body. This is calculated as late as
possible by calling _handle_body(), and this is all done privately
within the module. Therefore, it's easiest if this is done in a
centralised place as late as possible by this module.

Every attempt I made at implementing this by extended the relevant
code just resulted in oodles of unnecessarily duplicated code compared
to what is possible by just integrating it directly.

I know this was previously tackled six years ago in #70. In that PR,
the concern was that the header was being added unconditionally. I
have mitigated this problem in my PR by controlling the new behaviour
with a flag. There are still plenty of situations where you wouldn't
want the Content-Length header to be set, so this new behaviour is
fully backwards-compatible.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -387,6 +387,7 @@ class Response(BaseResponse):
         headers=None,
         stream=False,
         content_type=UNSET,
+        auto_calculate_content_length=False,
         **kwargs
     ):
         # if we were passed a `json` argument,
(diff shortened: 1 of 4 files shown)


responses/__init__.py
```diff
@@ -417,6 +419,13 @@ def get_response(self, request):
         headers = self.get_headers()
         status = self.status
         body = _handle_body(self.body)
+
+        if self.auto_calculate_content_length and \
```
beliaev-maksim, line 423: I would recommend to use parenthesis () instead of \

djmattyg007:
@markstory @beliaev-maksim I've updated this PR with code style changes recommended by black.

PR #400, opened 2021-09-10
Feature Passthrough Flag

### LINKED ISSUE #399: add a passthrough flag to BaseResponse
I am currently using *responses* to mock an API for running tests during development. However, I want to reuse these tests when sending the requests to the actual API, where I also want to use the `assert_call_count` method.
A solution could be to add the endpoints as passthru instead of mocked responses, but this would require to add the `passthru_prefixes` to the summation of `assert_call_count`.
Another solution could be a flag in the `BaseResponse` class which allows to activate a pass through callback for the matching request if the flag is set, otherwise use the predefined response. This would also make the `passthru_prefixes` obsolete, because they could be implemented as normal `Response` objects where this flag is set to `True`, so the matching logic of the normal Response instances could be reused.

I think this is somehow related to #388, because such a "passthrough on condition" as in the given example could be implemented with a `passthrough` property (or getter method) in `BaseResponse` which could be checked in `RequestsMock._on_request` to decide if the incoming request should be sent to the actual endpoint or if the response should be build from the mock data.

Many thanks for building this library. Looking forward for some feedback to the proposed feature.

carloshorn:
This PR closes #399 by adding a `passthrough` flag to `BaseResponse`, which can be used to let requests pass through instead of returning the defined mock values.

It may be discussed, in how far it also closes other issues mentioned in #399.


responses/__init__.py
```diff
@@ -598,6 +604,10 @@ def add_passthru(self, prefix):
 
         >>> responses.add_passthru(re.compile('https://example.com/\\w+'))
         """
+        warnings.warn(
+            "This feature is deprecated in favor of using a PassthroughResponse",
```
markstory, line 608: The proposed feature can't replace this method as many applications of `add_passthru` are not defining individual responses. Instead they allow all traffic for a specific host or path to pass through.

There might be a solution where `add_passthru` creates passthrough responses. That would be more backwards compatible and allow the definition of specific mocks that also passthrough traffic.

carloshorn: Good point. Here, `add_passthru` could simply create `PassThroughResponse` instances. The only problem for backward compatibility that I see is the method argument of the `BaseResponse` class, because there is no `ANY` method, but the current behavior of `add_passthru` would require it. I could imagine to implement this by using an `enum.IntFlag` so the response methods could be or-ed, e.g. `responses.PUT | responses.PATCH`. Then `ANY` would be the union of all methods. However, I need to investigate first how this works using older versions of python, because `enum.IntFlag` came with python 3.6.

carloshorn:
I pushed a version which should be backward compatible using `PassthroughResponse` objects instead of the `passthru_prefixes`.
In order to recover the behavior of the old `add_passthru` method, I needed to add an `ANY` method, to allow these responses to match like before (I could have just add separate responses for all possible methods, but this solution appears more elegant).
Additionally, I needed to add a `check_fired` flag to `BaseResponse` which is set to `True` by default, but which I need to switch off for the previous `add_passthru` behavior. A bit pragmatic is the reuse of this flag to sort these responses to the end when checking for matches. I would even call it a week point of the current implementation, because there could be a situation, where someone does not care about if it is fired or not, but the order still matters. Maybe there should be a match priority level which should be used for sorting, but I did not want to add another parameter yet.
Finally, to also cover the case where someone directly sets the `passthru_prefixes` in the `RequestsMock` instance, I had to convert this attribute to a property and handle the `PassthroughResponse` creation in the setter. Here, the check if a prefix already exists is a bit week, because I think it does not work yet for a regex comparison, which would cause a mock recreation. However, this should not be harmful, just an unnecessary object creation.

I hope this PR did not yet grow into a monster which should be split into smaller PR to reduce the risk of unforeseen side effects.

carloshorn:
After thinking about side effects, I will remove the use of `IntFlag` for HTTP methods, because there are also the `CONNECT` and `TRACE` methods defined, and a server could also define custom methods. Therefore, it would require a more complex method registration. Since I only need the `ANY` case, I could simply implement it within the if condition for method matching.

carloshorn:
I rebased this PR to avoid merge conflicts and simplified the implementation of the `ANY` method.
I also introduced a `priority` attribute for the response objects which is used for sorting them before finding the first match.
The priority is implemented as an `enum.IntFlag`, so a user can also just use an integer value to have a custom evaluation order which is not necessarily in the order these responses are added.
Finally, I added deprecation warnings for the direct interaction with the `passthru_prefixes` property. I do not expect many people to use this property (former attribute) directly and I could imagine that it vanishes in a future release.
If you think there should be a way to query the registered `Passthrough` responses, I might implement something as read only property, but I currently do not see the necessity and would follow the good old practice "*if in doubt leave it out*".

responses/__init__.py
```diff
@@ -601,7 +620,35 @@ def add_passthru(self, prefix):
...
+        self.add(response)
+
+    @property
+    def passthru_prefixes(self):
+        warn(
+            "Direct access of passthru_prefixes is depricated. "
+            "Passthroughs are stored as PassthroughResponses.",
+            DeprecationWarning,
+        )
+        return self._passthru_prefixes
```
markstory, line 636: Couldn't prefixes be generated at runtime from the pass through responses? I also don't understand why this method needs to be deprecated.

responses/__init__.py
```diff
@@ -601,7 +620,35 @@ def add_passthru(self, prefix):
...
+
+    @passthru_prefixes.setter
+    def passthru_prefixes(self, prefixes):
+        warn(
+            "Setting passthru_prefixes directly is depricated. Please use add_passthru",
```
markstory, line 641: Again, don't know why this is necessary.

responses/__init__.py
carloshorn: I think this is a misunderstanding, I absolutely in favour of the method `add_passthru`, because after getting to know it better, I clearly see the benefit. In the current implementation, it will not through any warning, because it will only interact with the private attribute `_passthru_prefixes`. The desire to deprecate the public attribute/property `passthru_prefixes` (especially the setter) comes from thinking about PEP 20 (Zen of Python): "*There should be one-- and preferably only one --obvious way to do it.*". The method `add_passthru` is already the preferred setter method, so I would discourage to offer this alternative way and therefore remove it from the public interface. Of cause, the property offers the possibility to register multiple passthroughs at once, but a loop just adds one line.

Still, I could imagine the situation, where someone wants to check which passthroughs are already registered. which I addressed with the comment
> If you think there should be a way to query the registered Passthrough responses, I might implement something as read only property, but I currently do not see the necessity and would follow the good old practice "if in doubt leave it out".
Therefore, I could imagine to expose the registered passthroughs with a property (only the getter), which instead of the prefixes would provide the corresponding `PassthroughResponse` instances, which facilitates to use the response registry methods (remove, replace, upsert) with them.

responses/__init__.py
carloshorn: See above.

carloshorn:
Thinking further about the simple access to passthroughs, I could imagine to have a more general search engine. Maybe the method `RequestsMock.search(method=None, url=None, passthrough=None, priority=None, ...)` which finds responses in `self._matches` with the desired attributes, e.g.
```python
my_indexes, my_responses = requests.search(method="ANY", passthrough=True)
all_indexes, all_responses = requests.search()
```
The index would also allow to determine the matching order, and maybe we could add a position argument to `RequestsMock.add` to inserts at a desired position. 

However, these thoughts are out of scope for this PR and might be discussed in another feature request and PR.

carloshorn:
Alright, I soften the deprecation, because I agree that there is not yet an adequate replacement for direct access of passthroughs. Since `PendingDepricationWarning`s are ignored by the default warning filters, I guess nobody will notice, but it would be marked in the code. I would also volunteer for building a replacement as outlined in the previous comment.

carloshorn:
I noticed that there was no test for the `passthru_prefixes` setter and the constructor argument. Even though, I want to deprecate them, I clearly need to test the introduced code.

carloshorn:
@markstory, the PR is ready for the next review. I added the missing tests for all lines added and also restored the backward compatibility for a little problem I found.
Many thanks in advance!

carloshorn:
> [...] I could fix the backward compatibility by also checking the priority before popping something.

Looking at other PRs, I noticed that this PR could solve the problem in #311, by putting theose responses to different priority groups.

markstory:
I'm not sure the complexity that priority groups would introduce is required when mocks can be defined in the required order.

carloshorn:
I agree with worrying about introduced complexity which could still contain more side effects popping up in the future, especially because most of it is introduced because of backward compatibility.

However, I still think the concept of response objects defining allowed requests, where passing them through is a possibility to handle it, is beneficial.

The essential part of this PR is simply the `BaseResponse.passthrough` flag and the simple lines
```python
if match.passthrough:
    logger.info("request.allowed-passthrough", extra={"url": request.url})
    return _real_send(adapter, request, **kwargs)
```
And I would also propose to keep the `PassthroughResponse` class which is simple enough to understand and use.

The headache started when trying to use the `PassthroughResponse` also for the `passthru_prefixes` which are conceptually something different and I was not really aware of it at the beginning. 
These `passthru_prefixes` are a sort of fallback, which therefore required the `BaseResponse.priority` to move them to the end of the matching order, because a user can define them before registering any response. Furthermore, they required the `BaseResponse.check_fired` flag, because for `passthru_prefixes`, it is not checked if they got fired or not, and they also needed the `ANY` method, because technically these `passthru_prefixes` are only URL based, but do not further specify the incoming request (maybe the greatest pro argument for this feature).
Finally, registering `passthru_prefixes` as normal responses had the side effect of responses getting popped from the registry if a prefix would also match.
It is obvious that the use cases covered by `passthru_prefixes` can also be implemented with `PassthroughResponse`s, because at the end of the day, passthroughs are also just requests fired by the test. However, I understand that changing the library API is a very critical step, and therefore not an option.

So how about reverting any non-essential parts and pointing on this duality in the documentation?

carloshorn:
I rebased with master to avoid merge conflicts and I reverted to the essential parts as described in previous comment.
The only additional thing I left is the test for `passthru_prefixes` set in the `RequestsMock` constructor and setting them directly as attribute, because this part of the public interface has not been covered by any test yet.

markstory:
Thanks for taking the time to put this together and compromising with a more minimal solution.

PR #401, opened 2021-09-13
Fix names and path to matchers in changelog

michael-k:
`query_param_matcher` does not contain `s`.  But `json_params_matcher` and `urlencoded_params_matcher` do.


markstory:
Thank you 👍

PR #403, opened 2021-09-15
mbeliaev/kwargs check

### LINKED ISSUE #228: Allow to specify stream=True for callback responses

beliaev-maksim:
added check for requests and responses arguments comparison
updated add_callback to comply with all available parameters
added tests
added docstrings
added kwargs to add_callback

Closes #228


responses/__init__.py
```diff
@@ -656,7 +682,15 @@ def upsert(self, method_or_response=None, url=None, body="", *args, **kwargs):
...
 
     def add_callback(
-        self, method, url, callback, match_querystring=False, content_type="text/plain"
+        self,
+        method,
+        url,
+        callback,
+        match_querystring=False,
+        content_type="text/plain",
+        stream=False,
```
markstory, line 691: I didn't think callback responses had streaming.

responses/__init__.py
```diff
@@ -691,12 +728,20 @@ def __exit__(self, type, value, traceback):
     def activate(self, func):
         return get_wrapped(func, self)
 
-    def _find_match(self, request):
+    def _find_match(self, request, kwargs):
+        """
+        Iterates through all available matches and valites if any of them matches request
```
markstory, line 733:
```suggestion
        Iterates through all available matches and validates if any of them matches the request
```

responses/__init__.py
```diff
@@ -341,6 +344,25 @@ def _req_attr_matches(match, request):
...
+        :param kwargs: (dict) values provided to requests method
+        :return:
+            (bool) True if all args match
+            (str) reason with info which args failed
+        """
+        for arg in kwargs:
+            if hasattr(self, arg):
+                response_arg_val = getattr(self, arg)
+                requests_arg_val = kwargs[arg]
+                if response_arg_val != requests_arg_val:
```
markstory, line 359: This seems like it could cause existing mocks to start failing if the requests for those mocks don't have all the same arguments. Should we provide this as an opt in matcher like `responses.matchers.all_kwargs_matcher()` instead?

beliaev-maksim: idea here is to really match the request that was requested. Otherwise, you may expect that you request `stream=True` but in fact you get no stream.
in such a case we will raise, this will be applied only to attributes that are assigned on the Response object, otherwise they are passed through

I can change the behavior to use matchers, but then I would need to change signature of the `match()` function to take kwargs as well
OR
we attach kwargs to request object before it is gone for validation, I need to check `requests` module how they handle kwargs

so, which option will you prefer: 1, 2, or 3 ? :D

responses/__init__.py
```diff
@@ -260,14 +260,17 @@ class BaseResponse(object):
...
+    def __init__(self, method, url, match_querystring=_unspecified, match=[], **kwargs):
         self.method = method
         # ensure the url has a default path set if the url is a string
         self.url = _ensure_url_default_path(url)
         self.match_querystring = self._should_match_querystring(match_querystring)
         self.match = match
         self.call_count = 0
 
+        for arg, val in kwargs.items():
+            setattr(self, arg, val)
```
markstory, line 272: Could we set the kwargs to a `request_kwargs` or `extra_args` attribute instead of setting them on the response? Splatting all the kwargs on to the response could limit us in the future as we can't add new properties or methods without potentially breaking userland code.

responses/__init__.py
markstory: Attaching the kwargs to the simulated request and using a matcher would be preferrable. We're already mutating the request to add `request.params` and the `match` API allows us to keep the logic for kwarg matching in one place, and makes enabling this feature explicit and straightforward to the reader of test code.

beliaev-maksim:
@markstory 
sorry for delay, finally got time to work on it

hope this PR will bring more user friendly messaging into matchers + new matcher

responses/__init__.py
beliaev-maksim: I decided not to touch stream in this PR

responses/__init__.py
beliaev-maksim: done

responses/__init__.py
beliaev-maksim: removed and moved to matchers

responses/test_responses.py
```diff
@@ -1527,76 +1568,106 @@ def run():
...
+    """
 
-        responses.add(
-            method=responses.GET,
-            url="http://111.com/",
-            body="two",
-            match=[matchers.json_params_matcher({"page": "one"})],
-        )
+    def run():
+        with responses.RequestsMock(assert_all_requests_are_fired=False) as rsps:
```
markstory, line 1603: Why did you move the tests to using `responses.RequestsMock()` over the module level methods?

beliaev-maksim: @markstory 
I wanted to limit the context.
Do you see any drawbacks of doing that?

beliaev-maksim: also, in this particular change I also separated tests that raise from matchers and from responses themselves.

markstory: I don't think there are significant drawbacks to using `RequestsMock()`. The only one I can think of is that tests have multiple styles of using the library and it could be harder for future contributors and users to know which interface is the intended usage.

beliaev-maksim: if strongly required, I can do a change for response decorator

responses/matchers.py
```diff
@@ -12,6 +12,18 @@
...
 
 
+def _create_key_val_str(input_dict):
+    """
+    Returns string of format {'key': val, 'key2': val2}
+    :param input_dict: dictionary to transform
+    :return: (str) reformatted string
+    """
+    key_val_str = "{{{}}}".format(
+        ", ".join(["{}={}".format(key, val) for key, val in sorted(input_dict.items())])
```
markstory, line 22: Should this recurse into values that are not scalars? or will native python formatting be ok?

responses/test_responses.py
markstory: Its not a strong requirement :smile:

responses/matchers.py
beliaev-maksim: nice point, I added a recursive conversion. Now it is a Frankenstein instead of nice one liner :D

now message will be truly compatible between py2 and py3. But please, have a look at implementation, if needed, I will revert commit back

beliaev-maksim: @markstory, if got some time, can you please review

beliaev-maksim:
@markstory 
if you can review, would be great. I have some ideas of further module changes, but they will be built on top of this PR

responses/matchers.py
beliaev-maksim: let's just use `json.dumps`, I assume in none of the cases key in dict will be an integer, so we can use that method with `sort_keys=True`

README.rst
```diff
@@ -129,7 +129,16 @@ auto_calculate_content_length (``bool``)
...
 
 match (``list``)
-    A list of callbacks to match requests based on request body contents.
+    A list of callbacks to match requests based on request attributes.
+    Current module provides multiple matchers that you can use:
+
+    * body contents in JSON format
+    * body contents in URL encoded data format
+    * request query parameters
+    * kwargs provided to request eg stream, verify
```
markstory, line 138:
```suggestion
    * kwargs provided to request e.g. ``stream``, ``verify``
```

responses/__init__.py
```diff
@@ -682,6 +682,14 @@ def activate(self, func):
         return get_wrapped(func, self)
 
     def _find_match(self, request):
+        """
+        Iterates through all available matches and validates if any of them matches the request
+        :param request: (PreparedRequest), request object
```
markstory, line 687:
```suggestion
        Iterates through all available matches and validates if any of them matches the request

        :param request: (PreparedRequest), request object
```

responses/matchers.py
```diff
@@ -12,6 +12,44 @@
...
+    """
+    Returns string of format {'key': val, 'key2': val2}
+    Function is called recursively for nested dictionaries
+    :param input_dict: dictionary to transform
+    :return: (str) reformatted string
+    """
+
+    def list_to_str(input_list):
+        """
+        Covert all list items to string.
```
markstory, line 25:
```suggestion
        Convert all list items to string.
```

responses/matchers.py
```diff
@@ -12,6 +12,44 @@
     JSONDecodeError = ValueError
 
 
+def _create_key_val_str(input_dict):
+    """
+    Returns string of format {'key': val, 'key2': val2}
+    Function is called recursively for nested dictionaries
+    :param input_dict: dictionary to transform
```
markstory, line 19:
```suggestion
    Function is called recursively for nested dictionaries

    :param input_dict: dictionary to transform
```

responses/matchers.py
```diff
@@ -69,19 +115,53 @@ def query_param_matcher(params):
...
+
+        return valid, reason
+
+    return match
+
+
+def request_kwargs_matcher(kwargs):
+    """
+    Matcher to match keyword arguments provided to request
+    :param kwargs: (dict), keyword arguments, same as provided to request
```
markstory, line 142:
```suggestion
    Matcher to match keyword arguments provided to request

    :param kwargs: (dict), keyword arguments, same as provided to request
```

PR #407, opened 2021-09-19
add matcher for request headers

eviltwin:
I needed a matcher for request headers and so I built one that I felt was generic enough to warrant inclusion upstream.

Hopefully the detail that I've added to the README.rst is enough to understand the motivation for how I built it, so I'll avoid repeating myself here.

I'm happy to discuss any changes that you might like in order to have this merged.


responses/matchers.py
```diff
@@ -85,3 +85,32 @@ def match(request):
...
+        request_headers = request.headers or {}
+
+        if allow_additional:
+            # filter down to just the headers specified in the matcher
+            request_headers = {k: v for k, v in request_headers.items() if k in headers}
+
+        valid = sorted(headers.items()) == sorted(request_headers.items())
+
+        if not valid:
+            return False, "%s doesn't match %s" % (
```
markstory, line 109:
```suggestion
            return False, "headers of %s doesn't match %s" % (
```

This might help make it more clear that headers were the source of the match failure.

markstory (commented):
Nice work. I only had one small suggestion.

responses/matchers.py
eviltwin: Fair, I can make that change. In practice you see messages like `"[('Authorization', 'foo')] doesn't match [('Authorization', 'bar')]"` so it's usually pretty obvious, but I have no qualms making it more obvious :)

I'll update the PR this evening

eviltwin:
Looks like I messed up on the python 2.7 messages - I'll be honest and admit that I typed them manually and didn't bother checking they were right... I'll update them along with the changes requested this evening.

responses/matchers.py
markstory: I agree that it *should* be pretty simple, but if header matching is combined with parameters matching it could be ambiguous and a few extra words could save a developer some time down the road.

eviltwin: Also, sorry for saying "this evening" 26 days ago 😳 that's bad time-management even for me 🤦‍♂️

responses/matchers.py
```diff
@@ -77,11 +87,40 @@ def match(request):
...
         return valid, ""
 
     return match
+
+
+def header_matcher(headers, allow_additional=True):
+    """
+    Matcher to match 'headers' argument in request using the responses library.
+    :param headers: (dict), same as provided to request
+    :param allow_additional: (bool), ignore and allow additional headers that
```
beliaev-maksim, line 104: may be call argument `strict_match` and set default to `False`
may be it is just my personal, but for me that sounds more explicit and easier to understand

then you check
~~~python
if not strict_match:
    for header in headers:
        if header not in request.headers:
            return False
~~~


~~~python
if strict_match:
    sorted(headers) == sorted(request.headers)
~~~

README.rst
```diff
@@ -193,6 +193,90 @@ Note, you must set ``match_querystring=False``
...
+        resp = requests.get("http://example.com/", headers={"Accept": "application/json"})
+        assert resp.json() == {"content": "hello world"}
+
+        resp = requests.get("http://example.com/", headers={"Accept": "text/plain"})
+        assert resp.text == "hello world"
+
+Because ``requests`` will send several standard headers in addition to what was
+specified by your code, request headers that are additional to the ones
+pased to the matcher are ignored by default. You can change this behaviour by
+passing ``allow_additional=False`` to the matcher to ensure that only the
```
beliaev-maksim, line 240: if decided, then rename arg here as well

responses/matchers.py
eviltwin: I really struggled to come up with a good name here, and I'd agree that yours is better. I'll make this change.

eviltwin:
thanks to both @beliaev-maksim and @markstory for your input 🙂

I've rebased this against master, changed the param name and squashed things together to tidy up the history. Hopefully it now incorporates everything discussed 👍

beliaev-maksim:
code implementation looks good to me

PR #409, opened 2021-09-27
Add passthrough responses to CallList and increment call count

### LINKED ISSUE #408: Passthrough Responses should appear in CallList
Currently, Passthrough responses do not appear in the `CallList` after termination of the test. A successful match does also not increase the call count of a `PassthroughResponse`.

carloshorn:
This PR closes #408.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -771,8 +771,11 @@ class RequestsMock(object):
                 raise
 
         if not match.stream:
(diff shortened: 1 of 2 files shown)


carloshorn:
I encountered a problem when `passthrough_response.stream = False`, but the request is sent as stream.

markstory:
Thank you 👍

PR #410, opened 2021-09-28
clean up unused function

beliaev-maksim:
remove functioned that was designed to include redirect resolving logic from requests.sessions.Session.

Now this logic is not used anymore in the code. For details see #63


markstory:
Thank you 👍

PR #411, opened 2021-10-05
refactor find matches algorithm

carloshorn:
This PR changes the algorithm to find matches. It implements the following algorithm:

Loop in normal order and provide responses with `call_count == 0`. If there is no match found, loop in reverse order and provide all responses with `call_count > 0`.

This will behave like before for most use cases. There could only be cases that failed before which would now pass, e.g. imagine the following situation:
`registry: [r'.*', 'abc']`.
In previous implementation a call with
`requests: ['abc', 'abc', 'xyz']` would have failed, because the first `abc` would have popped the "match everything" response. In this implementation, it would still work.

In my opinion this algorithm is easier to explain and comprehend and should be backward compatible with all previous use cases. Except those relying on a failure due to a popped response, which would be a very special edge case which I would not expect by any user at all.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -680,6 +680,10 @@ class RequestsMock(object):
         return self._matches
 
     @property
+    def registry(self):
+        return self._matches
+
+    @property
     def calls(self):
         return self._calls
 
@@ -696,22 +700,24 @@ class RequestsMock(object):
     def activate(self, func):
         return get_wrapped(func, self)
 
+    def _matching_sequence(self):
+        for response in self.registry:
+            if response.call_count == 0:
+                yield response
+        for response in reversed(self.registry):
+            if response.call_count > 0:
+                yield response
+
     def _find_match(self, request):
-        found = None
-        found_match = None
-        match_failed_reasons = []
-        for i, match in enumerate(self._matches):
-            match_result, reason = match.matches(request)
-            if match_result:
-                if found is None:
-                    found = i
-                    found_match = match
-                else:
-                    # Multiple matches found.  Remove & return the first match.
-                    return self._matches.pop(found), match_failed_reasons
-            else:
-                match_failed_reasons.append(reason)
-        return found_match, match_failed_reasons
+        match = None
+        fails = []
+        for response in self._matching_sequence():
+            success, reason = response.matches(request)
+            if success:
+                match = response
+                break
+            fails.append((response, reason))
+        return match, fails
 
     def _parse_request_params(self, url):
         params = {}
@@ -746,10 +752,8 @@ class RequestsMock(object):
                 "- %s %s\n\n"
                 "Available matches:\n" % (request.method, request.url)
             )
-            for i, m in enumerate(self._matches):
-                error_msg += "- {} {} {}\n".format(
-                    m.method, m.url, match_failed_reasons[i]
-                )
+            for m, reason in match_failed_reasons:
+                error_msg += "- {} {} {}\n".format(m.method, m.url, reason)
 
             response = ConnectionError(error_msg)
             response.request = request
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -640,9 +640,10 @@ class RequestsMock(object):
         else:
             response = Response(method=method_or_response, url=url, body=body, **kwargs)
 
-        try:
-            index = self._matches.index(response)
-        except ValueError:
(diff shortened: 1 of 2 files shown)


responses/__init__.py
```diff
@@ -679,6 +680,10 @@ def add_callback(
     def registered(self):
         return self._matches
 
+    @property
+    def registry(self):
```
markstory, line 684: Why do we need this property? We already have `registered()`

responses/__init__.py
```diff
@@ -696,22 +701,27 @@ def __exit__(self, type, value, traceback):
...
+    def _matching_sequence(self):
+        index = 0
+        for response in self.registry:
+            if response.call_count == 0:
+                yield index, response
+            index += 1
+        for response in reversed(self.registry):
+            index -= 1
+            if response.call_count > 0:
+                yield index, response
```
markstory, line 713: Do we need to `yield` the index? It looks unused in all the places this method is called.

responses/__init__.py
```diff
@@ -640,9 +640,10 @@ def replace(self, method_or_response=None, url=None, body="", *args, **kwargs):
         else:
             response = Response(method=method_or_response, url=url, body=body, **kwargs)
 
-        try:
-            index = self._matches.index(response)
-        except ValueError:
+        for index, match in self._matching_sequence():
```
markstory, line 643: If this now sometimes loops in reverse isn't the doc string inaccurate as the first response isn't necessarily the one that will be replaced.

responses/__init__.py
carloshorn: I needed to add it for the `replace` method (see second commit), which replaces the next match in matching order. Therefore `_matches.index` (now giving the next index in insertion order) needed to be replaced and I thought yielding index together with the match would facilitate the localization without additional helper functions.

responses/__init__.py
carloshorn: Kind of inaccurate, because now we would need to differentiate between *first response* in insertion order and *first response* in matching order. I think that the matching order is more relevant for users in most cases.

This leads me to a question about the intended use of the public `registered` method.

The documentation says:
> Registered responses are available as a public method of the RequestMock instance. It is sometimes useful for debugging purposes to view the stack of registered responses which can be accessed via responses.registered().

Currently, the `registered` method gives direct access to the **mutable private attribute** `_matches`, which allows direct manipulation of the internal state of the `RequestsMock` instance. I know that there is not a list view object in python, but from the documentation, it appears to me like the user should only get a view by this method and manipulate the registry with the `add`, `replace`, `upsert`, and `remove` methods.
Such a view (even though static compared to `dict.values()` for instance) could be implemented by returning `_matches.copy()`, or `[response for index, response in self._matching_sequence()]`. I would vote for the second one, because as said I think matching order is more relevant to the user. I would also say that it is most backward compatible for the debugging use-case mentioned in the documentation, meaning that users would like to know why another match had been picked instead of the one they expected, so they want to see the matching order before and after doing their request.
Now the question: do you see any conflict or critical backward incompatibility by changing the return value of the `registered` method?

carloshorn:
Thinking further, I could also imagine to move all the registry logic into a `Registry` class providing an `__iter__` method which loops in matching order and an `index` method which returns the index of the first response in matching order. It would also provide the manipulation methods like `add`, `upsert`, etc. and of course a `clear` method for the `RequestsMock.reset` method.
This would allow for alternative implementations of the matching order, so users like in #311 could provide their own `Registry` implementation which for examples always loops in insertion order, or only yield responses which have not been called (i.e. not providing the last inserted match as fallback by reverse looping), or do whatever they need.
However, before providing an implementation in this PR, I would like to know if it is something you would welcome in this library.

responses/__init__.py
markstory: Ah I missed that as the usage was outside of the modified lines.

responses/__init__.py
markstory: > Currently, the registered method gives direct access to the mutable private attribute _matches, which allows direct manipulation of the internal state of the RequestsMock instance. I

I think giving a mutable reference isn't the end of the world given that this library's primary purpose is to help with testing. We could clone the list but that adds complexity and potential for breaking userland code where we don't need to incur the complexity or risk.

markstory:
Going to close this as #416 was merged.

PR #415, opened 2021-10-18
mbeliaev/query_string

beliaev-maksim:


hramezani:
I think you can add some documentation for this new feature to the readme file. like the doc that we have for `query_param_matcher`

PR #416, opened 2021-10-21
mbeliaev/registry

beliaev-maksim:
I went through #411 and, sincerely, I had absolutely the same need and was thinking how would be better to implement it.
thanks to @carloshorn, who inspired with the idea of Registry

I propose fully backwards compatible implementation.
Short description:
1. User is provided with `registries.BaseRegistry` as abstract class that must be implemented
2. in custom class use can define any logic of finding match (user just redefines `find` function that must return `found_match, match_failed_reasons`), eg examples mentioned in #411 
  * strict: provide every response only once.
  * very strict: only provide the next response.
  * insertion: always provide responses in insertion order.
  * rotating: move found matches to the end, i.e. if they are already in order it performs a rotation (might play well with the replace method if the other responses stay the same.)
  * fall-back: the current selection order  
3. then user can replace registry for default mock by running `responses.set_registry(CustomRegistry)` or just set as argument in `RequestsMock`

idea is the same as with `matchers`. Once we shift to `registries`, then we can populate them with some predefined, eg like in #411 or some from point 2 above. Then, users will have easy access to new types of response registries out of the box with single line of code

Note:
once we agreed to proceed with implementation, then I will polish this PR: docs, test, typing, changelog


responses/__init__.py
```diff
@@ -511,16 +512,27 @@ def __init__(
...
         self.response_callback = response_callback
         self.passthru_prefixes = tuple(passthru_prefixes)
         self.target = target
 
+    def get_registry(self):
+        return self._registry
+
+    def set_registry(self, new_registry):
+        self.reset()
+        self._registry = new_registry()
```
carloshorn, line 530: I think a change of the Registry class should not reset the registered responses. Instead, the new instance should be initialized with the previously registered responses.
You could easily do that by giving the `BaseRegistry` constructor an optional argument which sets the `_responses` attribute and make it itterable, so you could easily switch between implementations.

carloshorn:
I had to implement my own version of this feature to understand your design choices regarding the setter and getter for the registry. So you need the setter method for interaction with the module level default instance of `RequestsMock`, which you achieve by exporting the setter function.

I agree with @markstory that *custom registries is only for niche/power user usage*. So I implemented the registry as normal attribute (not as property), because I think those *niche/power users* will likely use their own instance of `RequestsMock`. Then it will be simple to exchange the registry like any other attribute. Changing the default mock is dangerous as shown with the leaking `CrashRegistry` above. So I would suggest to only allow those things on self created instances.

@beliaev-maksim, I thank you very much for starting this, so I do not want to take your credits and start my own PR on this. Have a look at https://github.com/carloshorn/responses/tree/feature-registry to see my implementation proposal.

While thinking about module level properties, I wondered why `response_callback` is exposed. It will always be `None`, because it is just a reference to the default value. Changing registry is somehow related to changing the response_callback. It is something which should not be done on the default mock, because it could leak into other tests. Global objects can easily become dangerous.

CHANGES
```diff
@@ -1,3 +1,10 @@
+0.16.0
+------
+
+* Added `responses.registries`. Now user can create custom registries to
+  manipulate the order of responses in the match algorithm
+* `responses.activate` decorator should be used as `responses.activate()`
```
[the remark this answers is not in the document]
beliaev-maksim: will not deprecate, agree, just allow both

README.rst
```diff
@@ -353,6 +353,42 @@ include any additional headers.
...
+
+
+    class CustomRegistry(registries.FirstMatchRegistry):
+        pass
+
+
+    # using function decorator
+    @responses.activate
+    def run():
+        responses.set_registry(CustomRegistry)
```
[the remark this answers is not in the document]
beliaev-maksim: missed it

responses/__init__.py
```diff
@@ -691,8 +705,18 @@ def __exit__(self, type, value, traceback):
         self.reset()
         return success
 
-    def activate(self, func):
-        return get_wrapped(func, self)
+    def activate(self, func=None, custom_registry=None):
+        if func is not None:
+            warn(
+                "Decorator 'activate' is deprecated. Use 'activate()' instead",
+                DeprecationWarning,
```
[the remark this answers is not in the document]
beliaev-maksim: will not deprecate

responses/__init__.py
```diff
@@ -888,3 +898,5 @@ def assert_call_count(self, url, count):
 stop = _default_mock.stop
 target = _default_mock.target
 upsert = _default_mock.upsert
+_set_registry = _default_mock._set_registry
```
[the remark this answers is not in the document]
beliaev-maksim: removed

beliaev-maksim:
@carloshorn 
you probably missed my last commits. After you made a post about Registry leakage, I re-implemented it. Now registry is only valid within a single test or inside of `RequestMock`

I just pushed updated docs, you can see that there is no leakage any more

beliaev-maksim:
@markstory 

sorry for the long thread here. But feature required some brainstorming :)
can you please review code and documentation again?

beliaev-maksim:
@markstory 
can you please review this PR?
I have some further changes that might depend on this PR

beliaev-maksim:
@markstory 
I resolved conflicts caused by recent updates to master
since 0.16 is scheduled, I bumped the version here to 0.17

as I mentioned, current PR is fully backwards compatible

README.rst
```diff
@@ -387,6 +387,55 @@ include any additional headers.
         resp = session.send(prepped)
         assert resp.text == "hello world"
 
+Response Registry
+---------------------------
+
+By default, ``responses`` module will search all registered``Response`` objects and
```
markstory, line 393:
```suggestion
By default, ``responses`` will search all registered``Response`` objects and
```

README.rst
```diff
@@ -387,6 +387,55 @@ include any additional headers.
...
+
+    import responses
+    from responses import registries
+
+
+    class CustomRegistry(registries.FirstMatchRegistry):
+        pass
+
+
+    print("Before tests:", responses.mock._get_registry())
```
markstory, line 415: I don't think we should be giving examples of using non-public examples. The direction taken in #440 where there is a public `.registry` would let users who need to interact with the registry use public APIs.

README.rst
```diff
@@ -387,6 +387,55 @@ include any additional headers.
...
+
+    # using function decorator
+    @responses.activate(registry=CustomRegistry)
```
markstory, line 418: Do you find that yo need to change the registry strategy in each test? Should we only support `responses.registry` as the way to change the registry globally and `RequestsMock.registry` for a locally scoped registry when it is required?

beliaev-maksim: @markstory 
the idea here is similar to `mock.patch` where you patch each test.

in general, we can compliment this implementation with making `get` and `set` registry public methods. Exposing `registry` directly allows to set non empty registry. The question would be if we want to allow this or do we want to be more conservative and allow only empty registry that should be filled?

README.rst
beliaev-maksim: updated

README.rst
beliaev-maksim: let me provide a detailed example.

one of the worst things in unit testing that can happen, is a leakage of environment from one test to another
Let's take an example, based on proposed alternative implementation where one gets `.registry` attribute:
```python
import responses
from responses.registries import StrictOrderRegistry, PopFirstKeepLastRegistry

assert type(responses.mock.registry) == PopFirstKeepLastRegistry

    rsps.registry = StrictOrderRegistry()
    assert type(rsps.registry) == StrictOrderRegistry


@responses.activate
def my_test_func():
    responses.mock.registry = StrictOrderRegistry()
    assert type(responses.mock.registry) == StrictOrderRegistry


my_test_func()
assert type(responses.mock.registry) == PopFirstKeepLastRegistry
# >>> AssertionError, since responses.mock.registry is equal to StrictOrderRegistry
```
what we see, that if anybody in some test will assign value of the registry, then it will propagate to all other tests.
we need to support identical behavior for both Context Manager and for decorator. Of course, we can revert to default registry on `__exit__`, but this will be:
1. implicit (I prefer to be explicit whenever it is possible) 
2. will cause issues when `responses.mock.registry` set outside of the function (globally), since this will be applied only to a first function executed

on top of this we have additional minor issue, user can make a typo in attribute name and still think that registry is set, while it is not. eg `responses.mock.regstry`

this leads that we need to set registry on each test, reasons for this:
1. explicit
2. allows user to precisely control behavior of tests

drawback that we need to do extra coding is eliminated with new proposal of **backwards compatible** decorator, that can take argument of registry `@responses.activate(registry=CustomRegistry)` or `registry` argument in context manager itself. Both ways should be easy to use.

carloshorn: > [...] Let's take an example, based on proposed alternative implementation [...]

This is the point where I think I need to explain the intention and difference between this PR and #440.
In #440 the implementation is not advertising the registry to be used on the module level instance `responses.mock`. Instead it is implemented similar to `responses.mock.response_callback`, where all the warnings about the leakage of environment are equally valid for the current version of this library. However, in contrast to the module level `response_callback` which is just a reference to `None`, a module level reference to `registry` could make sense, assuming that the registry of the default mock is never swapped out.
So the difference is, that this PR tries to implement the registry functionality to be also used with the default mock, while #440 intents to provide the functionality only for local instances of `RequestsMock`.
Maybe it is just me who is scared about global (module level) instances, but note how much complexity is needed to restore the initial state of the default mock instance... What about a global reset function which just recreates a new instance of default mock with a well defined state? Of course, this reset function will also need to reset the global pointers to the new instance methods and attributes.

PR #423, opened 2021-10-27
Fix regression from stream deprecation

### LINKED ISSUE #421: Cookie handling regression from 0.14.0 to 0.15.0
### Environment

Using the responses library under Python 3.9.7

### Steps to Reproduce

Run this test against responses 0.14.0 and against responses 0.15.0

```
#!/usr/bin/env python3
import sys
import requests
import responses



def main():
    with responses.RequestsMock() as fakes:
        fakes.add(responses.GET, 'https://example.com', status=200,
            adding_headers=[
                  ('Set-Cookie', 'mycookie=myvalue; path=/; secure'),
            ],
        body="Returned body",
        )

        session = requests.session()
        response = session.get('https://example.com')

        tests = {}
        tests['response-cookie'] = 'mycookie' in response.cookies
        tests['session-cookie'] = 'mycookie' in session.cookies

        failures = 0
        for testname, result in tests.items():
            if not result:
                failures += 1
            sys.stdout.write(f"Test {testname} result {result}\n")
        if failures:
            sys.stdout.write(f"{failures} of {len(tests)} failed.\n")
        else:
            sys.stdout.write(f"All {len(tests)} tests passed.\n")

        return 1 if failures else 0

if __name__ == '__main__':
    sys.exit(main())
```

### Expected Result
Expected 'mycookie' to appear in both the response.cookies and the session.cookies

### Actual Result

'mycookie' appears in both the response.cookies and the session.cookies for 0.14.0
'mycookie' appears in the response.cookies *but not* in the session.cookies for 0.15.0

retracile:
The stream handling seems to break things; and it appears that the response is already in the correct state anyway.

Fixes #421

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -783,7 +783,10 @@ class RequestsMock(object):
         stream = kwargs.get("stream") if match.stream is None else match.stream
         if not stream:
             content = response.content
-            response.raw = BufferIO(content)
+            if kwargs.get("stream"):
+                response.raw = BufferIO(content)
+            else:
+                response.close()
 
         response = resp_callback(response) if resp_callback else response
         match.call_count += 1
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -780,14 +780,6 @@ class RequestsMock(object):
                 response = resp_callback(response) if resp_callback else response
                 raise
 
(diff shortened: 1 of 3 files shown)


markstory:
I fixed the conflict here and squash merged these changes into 4974c853a7babb9f260728b938af8565259d7d22

PR #424, opened 2021-10-30
mbeliaev/coverage

beliaev-maksim:
* increase coverage
* remove part with omitting self that was not covered. We have tests that cover when decorator is set on a method of a class + added tests for pytest.fixtures to ensure stability


markstory:
Thank you 👍

PR #427, opened 2021-11-01
mbeliaev/identifier

beliaev-maksim:
same as #108 but,
since by default identifier is ignored by `responses` we just use existing mechanism with `matchers`

this helps us not to expose additional argument in `Response` and follows logic of the rest of the library


markstory:
Thank you 👍

PR #429, opened 2021-11-01
update CI

beliaev-maksim:
run coverage on `responses` module only. Same as done in tox.ini

basically, this will exclude `setup.py` from coverage report


markstory:
Thank you 👍

PR #432, opened 2021-11-04
mbeliaev/query match

beliaev-maksim:
Deprecate `match_querystring`

Current behavior  (_Include the query string when matching requests.
    Enabled by default if the response URL contains a query string,
    disabled if it doesn't or the URL is a regular expression._) might be not intuitive

* It is better to explicitly specify `matcher` with requested params.
* Change simplifies `Response` signature and lets `matchers` do its job and match URL
* This change also simplifies library code


beliaev-maksim:
@markstory 
merge conflict is resolved

PR #434, opened 2021-11-05
Move tests for matchers into a separate file

markstory:
The test_responses file is getting big, and this separate mirrors how the code is separated. I didn't change the tests as I think having integration tests are important as that's how users use responses.


beliaev-maksim:
@markstory 
I had the same idea ahaha

PR #436, opened 2021-11-11
fix MANIFEST.in

### LINKED ISSUE #38: Provide the test data in the PyPI tarball
Hi,

Could you please provide the data used for the tests in the tarball next time you upload to PyPI? I'm packaging responses for Debian, and would very much like to be able to run the tests when building the package.

Thanks in advance!

beliaev-maksim:
closes #38

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 include README.rst CHANGES LICENSE
-include test_responses.py tox.ini
+include test_responses.py test_matchers.py
 include **/*.pyi
+include tox.ini
 global-exclude *~


beliaev-maksim:
@markstory
This one is a must have for 0.16

markstory:
I don't think it will make it in for 0.16 as the release is already scheduled, but I don't think missing tests from the wheel is a deal breaker.

beliaev-maksim:
@markstory
Yea, I noticed too late that test was not in manifest. Was checking old issues and found this one

Not critical, but still better to have consistency across packages

Issue #376, opened 2021-03-17
Error or better document overlapping regex matcher behaviour 


markddavidoff:
### Alternate names for this issue
Alternate name: "raise error instead of silently popping when multiple matches found (or add to readme)"
Alternate name2:  remove the `# Multiple matches found.  Remove & return the first match.` in `_find_match()`

### <3
First of all, I <3 responses. SOOOO extensible and amazing, Thank You!

###  Issue
I had a huge headache using multiple callbacks with different (overlapping) regexes. I assumed that the order of the regex matches mattered so i could do something like first add specific regex callback, then add general `.*` regex callback and the first one would match the specific case and then the general callback would be the "fallback" for the urls that didn't match the specific callback.

### Why I was confused
This made sense to me as regex is not super great at "not" matches, so i expected the implementation would allow for that. 

But instead I ended up with this weird behavior where the first match would trigger my specific callback and then the rest would trigger the general callback. And then at the end if i checked the registered matchers, I'd end up with one missing.

### Workaround
I was able to get this working by modifying my "general" regex to be the opposite (negative look ahead) of my specific regex so the general would not match the specific and there was no overlap.

### Issues with workaround
- With multiple specific cases the negative case in the general regex can get complicated.
- Regex negative lookarounds are expensive and hard to write.
- Also when you are adding callbacks in multiple different places having them not overlap can get complicated as well.

### Proposed Solutions
- Raise an error when we have multiple matching regexes instead of removing them. At the very least this should cause tests to fail
- Document this behavior explicitly in the readme

Jlrine2:
I think this is the expected behavior, if there is more than one match when a request is made, the first match is used for that request and then it is removed so that further request will use the later ones.

markddavidoff:
I can see that it was intentional, however, i would like it to be more
explicit enough if possible.

Silently removing callback mappings may cause unexpected side effects,
where the wrong callback was called.

In the case it is caused by poor regexes (which in my xp does happen) that
overlap, it would change which callback was called and in that case the
test might pass when it should not. I would rather the test error and fail
instead.

Jlrine2:
removing that code would completely break the way handling of multiple responses work though.

markddavidoff:
In your case for multiple responses I do find the existing code a bit odd though after looking at it more.

[...]

PR #440, opened 2021-11-13
responses registry - alternative implementation

carloshorn:
Last week, I ran again into a situation, where a custom registry could have made things easier. I had added my responses in order, and all I wanted was to pop one after the other. The problem was that some of the requests had queries and others not but they were sent to the same endpoint. I could fix it by setting `match_querystring`, but a strict pop one after the other registry would have made things more intuitive.

@beliaev-maksim has opened #416 to implement this feature which I proposed at the end of #411.
Here, I want to thank @beliaev-maksim for his effort in #416 and the fruitful discussions on this feature. However, I got the impression that his implementation might add complexity that keeps you from merging. So I decided to propose an alternative implementation which does not touch the public interface of the module and only focuses on the implementation of the registry. Intentionally, this implementation does not allow the usage of this feature on the libraries default mock, due to the risk of leaking into other tests as discussed in #416. Instead, I just added the registry attribute to the public interface of the `RequestsMock` class which can be accessed and replaced by advanced users on a local instance.
For convenience, I added the most relevant registry classes into the `registries` module. Those cover the current behavior, the always start matching in insertion without popping, and the strictly pop only the next response. These scenarios are also mentioned in #311 and #376. Optimizations as proposed in #411 are left to the users, which now do not need to hack into private methods and therefore fear incompatibility of their code with newer releases.

Example usage:

```python
import responses
from responses.registries import StrictOrderRegistry

    rsps.registry = StrictOrderRegistry()
    # add a single response for each request in requesting order
    ...
    resp.reset()
    # reset will just clear the registry which is fine, because it is only used in local instances
    # this is similar to using a response callback
    assert isinstance(rsps.registry, StrictOrderRegistry)
    assert len(rsps.registry) == 0
```

Further note that the Registry objects implement the `collections.abc.Collection` interface, which allows interactions like

```python
response in rsps.registry
for response in rsps.registry:  # or just iter(rsps.registry)
    ...
len(rsps.registry)
``` 

I could imagine to extend the interface to `collections.abc.MutableSequence`, so this object could become the return value of `RequestsMock.registered` without backward incompatibility, but in the current implementation, I just kept the previous behavior of giving excess to the private attribute `rsps.registry._responses`.
I could also imagine to add a `BaseRegistry`, but you already mentioned in #416, that you do not see necessity for it.


carloshorn:
Just noticed that the `StrictOrderRegistry` would also solve #180, because there would not be a fallback response remaining in the registry.

beliaev-maksim:
There is probably some misunderstanding, in #416 there is no leakage. Opposite to this PR, in #416 swapping of registry during test run is discouraged. User sets specific registry only for a single test or for a single context manager

carloshorn:
No offense @beliaev-maksim, I do not think there is any misunderstanding. Your PR implements an equivalent solution as I stated in the update of https://github.com/getsentry/responses/pull/416#issuecomment-955033144. However, your PR has not been merged yet, so there must be some concern about the implementation. 
The idea of this PR is to start with a minimalistic implementation regarding the change of any public interface. 
I know that you were very careful about not breaking backwards compatibility, but any additional keyword argument or additional method adds complexity which could be discussed as separate feature in a separate PR, e.g. the change of the module level activation decorator.
I am not saying that I am against your ideas or your implementation. I am just trying to get this feature into the library, where I can only guess on what is missing or too much in your proposal. At the end, we are working for the same goal and I am just trying a new attempt with a different approach.

markstory:
>  However, your PR has not been merged yet, so there must be some concern about the implementation.

I've been a bit short on time lately and haven't had a chance to go through this or #416 in depth yet as they are large changes. I'm hoping to get to that later this week.

markstory:
Going to close this as #416 was merged.

PR #442, opened 2021-11-15
chore(py2) Start removing python 2 support

markstory:
0.17 will start to remove support for Python 2. There is more work to be done to inline types and fully adopt py3 idioms but that can be done separately.

This first step removes python 2 from the CI matrix and removes most of the compatibility shim code.

### DIFF
--- a/.craft.yml
+++ b/.craft.yml
@@ -7,5 +7,5 @@ targets:
   - name: pypi
   - name: github
 requireNames:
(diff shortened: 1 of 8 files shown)


markstory:
I've added a deprecation for python 2.7 in #450

markstory:
Closing as #459 is more complete.

PR #453, opened 2021-12-01
docs: multipart_matcher argument order in example

ankush:
The signature of this function is:

https://github.com/getsentry/responses/blob/5799ff68ebc00a2d0faa605136862614cf3a2f88/responses/matchers.py#L234

However, in the README example, the order is reversed which can cause confusion.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -270,7 +270,7 @@ to the request:
         req_files = {"file_name": b"Old World!"}
         responses.add(
             responses.POST, url="http://httpbin.org/post",
-            match=[multipart_matcher(req_data, req_files)]
+            match=[multipart_matcher(req_files, data=req_data)]
         )
         resp = requests.post("http://httpbin.org/post", files={"file_name": b"New World!"})
 


README.rst
```diff
@@ -266,11 +266,11 @@ to the request:
 
     @responses.activate
     def my_func():
-        req_data = {"some": "other", "data": "fields"}
         req_files = {"file_name": b"Old World!"}
+        req_data = {"some": "other", "data": "fields"}
```
beliaev-maksim, line 270: what is the sense to move this line?

README.rst
```diff
@@ -266,11 +266,11 @@ to the request:
...
         responses.add(
             responses.POST, url="http://httpbin.org/post",
-            match=[multipart_matcher(req_data, req_files)]
+            match=[multipart_matcher(req_files, data=req_data)]
```
beliaev-maksim, line 273: this one is a good catch, thanks

beliaev-maksim (changes requested):

README.rst
ankush: Ah, nothing much. Only maintaining the order they appear in args. Can undo this if you want.

beliaev-maksim: less changes is always preferable, it allows to check PR easier and not to spend time understanding why change was done. Please undo this line

ankush: Done.

PR #456, opened 2021-12-02
mbeliaev/refactoring

beliaev-maksim:
some minor refactoring, fix types, doc strings, etc

PyCharm identified issues on latest master:
image

PyCharm identified issues after merging this PR:
image


responses/__init__.py
```diff
@@ -570,6 +570,8 @@ def __init__(
         self.response_callback = response_callback
         self.passthru_prefixes = tuple(passthru_prefixes)
         self.target = target
+        self._patcher = None
+        self._matches = []
```
markstory, line 574: Doesn't the reset() call on line 568 do this?

beliaev-maksim: @markstory 
yes, it does, but by python convention all class attributes should be initialized in `__init__` method

markstory: Ok. Couldn't we remove the call to `reset()` then?

beliaev-maksim: @markstory 
`reset` also resets calls:
`self._calls.reset()`

PR #457, opened 2021-12-31
Update ci.yml

beliaev-maksim:
see
https://github.com/codecov/codecov-action#%EF%B8%8F--deprecration-of-v1

UPD:
also, have to change pytest runner to generate XML file, I think this is related to codecov/codecov-action#577


beliaev-maksim:
work in-progress

looks like simple bumping does not work.
looks like related to https://github.com/codecov/codecov-action/issues/577

PR #458, opened 2022-01-07
mbeliaev/docs

beliaev-maksim:
docs grew a lot

makes sense to add TOC, see how it looks like below
also, updated contributor section

image


beliaev-maksim:
@markstory 
any reason why linter fails?
only docs were changed

also, I cannot reproduce it locally. I tried mypy 0.910 and 0.931

markstory:
>  any reason why linter fails?

Not sure. There aren't any errors on `master`. I'm fine with merging and if it fails on master I can get it fixed.

PR #462, opened 2022-01-10
mbeliaev/stubs for matchers

beliaev-maksim:
updated matchers.py with inline typing


responses/matchers.py
```diff
@@ -45,18 +47,18 @@ def list_to_str(input_list):
...
     :param params: (dict) data provided to 'data' arg of request
     :return: (func) matcher
     """
 
-    def match(request):
+    def match(request: PreparedRequest) -> Tuple[bool, str]:
         reason = ""
         request_body = request.body
-        qsl_body = dict(parse_qsl(request_body)) if request_body else {}
+        qsl_body = dict(parse_qsl(request_body)) if request_body else {}  # type: ignore[type-var]
```
beliaev-maksim, line 61: I do not think `mypy` is reasonable enough
for example here the fix could be 

```suggestion
        qsl_body = dict(parse_qsl(str(request_body))) if request_body else {}
```

however, I think this will break compatibility when body is `bytes`

and `mypy` errors:
```
responses/matchers.py:61:25: error: Value of type variable "AnyStr" of "parse_qsl" cannot be "Sequence[object]"  [type-var]
```

markstory: Yeah, I don't think the str() cast is always going to be safe. I would rather we put an ignore in than change the runtime code to satisfy a linter here.

PR #470, opened 2022-01-12
Fix 464

### LINKED ISSUE #464: `responses.add_callback` triggers DeprecationWarning beyond user's control
### Steps to Reproduce

My company uses responses in combination with pytest with deprecation warnings treated as failures. The dependabot upgrade from 0.16.0 to 0.17.0 failed.

This test (which I've scrubbed to avoid sharing irrelevant and company sensitive details) triggers a DeprecationWarning upon execution.

```python
    @responses.activate
    def test_something(self, caplog):
        # setup elided
    
        def request_callback(request):
            """Used to validate that the request correctly sets the status to deleted"""
            payload = json.loads(request.body)
            assert payload['StatusType'] == SOME_CONSTANT_TO_CHECK
            return 200, {}, json.dumps(SOME_RESPONSE_DATA)
    
       responses.add_callback(
            responses.POST, '/a/url/path', callback=request_callback
       )
```
Here is the warning:

```
E           DeprecationWarning: Argument 'match_querystring' is deprecated.
Use 'responses.matchers.query_param_matcher' or 'responses.matchers.query_string_matcher'
```

### Expected Result

This test makes no use of `match_querystring`. I would not expect to see a deprecation warning.

### Actual Result

A deprecation warning triggered.

I believe the reason is because `add_callback` sets `match_querystring` to `False` by default (see https://github.com/getsentry/responses/blob/0.17.0/responses/__init__.py#L719).

`_should_match_querystring`, the method where the deprecation warning is triggered, expects the more exact check of `if match_querystring_argument is not None:` (see https://github.com/getsentry/responses/blob/0.17.0/responses/__init__.py#L333)

My guess is that `_should_match_querystring` should be able to check the `False` value too to prevent disruption to the `add_callback` API.

Thanks for the help. Please let me know if I can do more to help or if more information is needed.

beliaev-maksim:
closes #464 

mock up with custom `False`

this allows to preserve deprecation warning and backwards compatibility with 0.16

### DIFF
--- a/responses/test_responses.py
+++ b/responses/test_responses.py
@@ -592,11 +592,11 @@ def test_callback_match_querystring_default_false():
         assert resp.status_code == status
         assert "foo" in resp.headers
 
-    from _pytest.outcomes import Failed
+    with pytest.warns(None) as record:
+        run()
 
-    with pytest.raises(Failed):
-        with pytest.deprecated_call():
-            run()
+    # check that no deprecation warning was raised
+    assert not record
 
     assert_reset()
 
--- a/responses/test_responses.py
+++ b/responses/test_responses.py
@@ -559,9 +559,46 @@ def test_callback():
     assert_reset()
 
 
-def test_callback_deprecated_argument():
+def test_callback_deprecated_stream_argument():
     with pytest.deprecated_call():
         CallbackResponse(responses.GET, "url", lambda x: x, stream=False)
+
+
+def test_callback_deprecated_match_querystring_argument():
+    with pytest.deprecated_call():
+        CallbackResponse(responses.GET, "url", lambda x: x, match_querystring=False)
+
+
+def test_callback_match_querystring_default_false():
+    """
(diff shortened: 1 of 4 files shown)


responses/test_responses.py
```diff
@@ -559,11 +559,48 @@ def run():
...
+
+    @responses.activate
+    def run():
+        responses.add_callback(responses.GET, url, request_callback, content_type=None)
+        resp = requests.get(url, params=params)
+        assert resp.text == "test callback"
+        assert resp.status_code == status
+        assert "foo" in resp.headers
+
+    from _pytest.outcomes import Failed
```
markstory, line 595: Is this a feature that is intended to be used? Another approach would be to use `with pytest.warns(None) as record:` and then assert that `record` is empty after calling `run()`.

beliaev-maksim: @markstory 
let me do a quick fix and check

beliaev-maksim: @markstory 
done

Issue #467, opened 2022-01-11
0.17.0 blocks moto and httpretty interop


davetrollope:
We have tests that utilize httpretty and moto but the 0.17.0 update causes the following error:

requests.exceptions.ConnectionError: Connection refused by Responses - the call doesn't match any registered mock.

This is true because our internal calls are not an AWS destination and are mocked by httpretty.

Pinning back to 0.16.0 resolved the issue

```
  File "/usr/local/lib/python3.6/site-packages/requests/api.py", line 76, in get
    return request('get', url, params=params, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/requests/api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/requests/sessions.py", line 530, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python3.6/site-packages/requests/sessions.py", line 643, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/responses/__init__.py", line 842, in unbound_on_send
    return self._on_request(adapter, request, *a, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/responses/__init__.py", line 821, in _on_request
    raise response
requests.exceptions.ConnectionError: Connection refused by Responses - the call doesn't match any registered mock.
```

markstory:
Are you using an pass through prefixes to exclude responses from attempting to handle specific requests?

markstory:
As a follow up is there a way to reproduce the failure you're having with 0.17 that passes with 0.16?

bblommers:
> Are you using an pass through prefixes to exclude responses from attempting to handle specific requests?

Moto maintainer here - yes, we do. This was our setup in 0.16:
```
responses_mock = responses.RequestsMock(assert_all_requests_are_fired=False)
responses_mock.add_passthru("http")
```

A reproducible usecase using Moto would be:
```
@moto.mock_sqs
def test_passthrough_requests():
    conn = boto3.client("sqs", region_name="us-west-1")
    conn.create_queue(QueueName="queue1")

    res = requests.get("https://httpbin.org/ip")
    assert res.status_code == 200
```
So the first request would be intercepted by `responses`, because our decorator has created a mock for it, but the second request should go through.

bblommers:
[...]

PR #471, opened 2022-01-12
Remove unused field

bblommers:
Per the discussion in #467 - this field is no longer used anywhere, so should be safe to remove.
All tests still pass after this change (locally..)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@
 * Dropped support of Python 2.7, 3.5, 3.6
 * Removed internal `_is_string` and `_ensure_str` functions
 * Removed internal `_quote` from `test_responses.py`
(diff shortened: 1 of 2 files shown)


CHANGES
```diff
@@ -4,6 +4,7 @@
 * Dropped support of Python 2.7, 3.5, 3.6
 * Removed internal `_is_string` and `_ensure_str` functions
 * Removed internal `_quote` from `test_responses.py`
+* Removed internal `_matches` from `__init__.py`
```
beliaev-maksim, line 7:
```suggestion
* Removed internal `_matches` attribute of `RequestsMock` object.
```

markstory:
Thank you 👍

PR #476, opened 2022-01-14
expose get_registry() method

### LINKED ISSUE #475: CustomRegistry - improve reset() availability
Hi all,

Thanks again for the CustomRegistry-implementation! It makes it for us at Moto much easier to have fine-grained control over the matching order/priority.

One painpoint I ran into, when converting everything to work with 0.1.7.0, was the `reset`-api. We have a global `RequestMock`-object that does need to be reset occasionally, but we want to keep the custom registry and the passthru's.

Our current implementation looks like this:
```
responses_mock.reset()
responses_mock._set_registry(CustomRegistry)
responses_mock.add_passthru("http")
```

There's two downsides to this: 
 - We're accessing the internal API
 - We're duplicating the configuration on creation and reset

### Some solutions that I've come up with:
1) Extend the reset-API to configure what needs to be reset, keeping the current implementation as default
```
def reset(keep_passthru=False, keep_registry=False):
    ...
```

2) Make the `set_registry(..)` API public

3) Make the `get_registry()` API public. This would allow us to call the `reset`-API of the registry directly:
```
def within_our_logic:
    responses_mock.get_registry().reset()
```

Anyone of these solutions would simplify our implementation, and make it less brittle, and would therefore be very welcome.
(All three of them would be even better of course, to have some flexibility!)

Any thoughts on whether these solutions are acceptable? I'd be happy to create a PR for them, just wanted to get some feedback first whether this would work for you.

beliaev-maksim:
closes #475 

@markstory 
we can consider backing up `_get_registry()` and complimenting it with `get_registry()` or just replacing the method

basically, it is an internal API now, so we can go straight with replacement, or we just do it extra safe


README.rst
```diff
@@ -482,20 +482,26 @@ Example that shows how to set custom registry
         pass
 
 
+    print("Before tests:", responses.mock.get_registry())
```
markstory, line 485: Can we get these removed?

beliaev-maksim: we can, but why?

I think this is a good example to show how to check registry type

markstory:
Looks good but there are some conflicts that need to be fixed up.

PR #477, opened 2022-01-18
mbeliaev/new_reg

### LINKED ISSUE #52: Support for chaining responses targeted at the same URL and method
I really love responses, but I had to start using HTTPretty, because I am testing against an API that always uses the same URL (ya I know....), so I have to chain the responses like this:

https://github.com/gabrielfalcao/HTTPretty#rotating-responses

Everything has been fine, but I realized that `match_querystring` does not do what you would expect. In my case, I have two of the same URL with different query strings that get called multiple times. Everything is fine on the first attempt of each, but after that, HTTPretty will just return the last result over and over (regardless of the query string).

I was wondering if responses is thinking about adding support for chaining responses (since query string matching seems to work properly in this module). I am going to work on adding this functionality to either HTTPretty or responses, and wanted to see if you guys would be interested in having such functionality.

Thanks!

### LINKED ISSUE #99: Feature request: response depending on invocation index
It could be interesting to be able to program different response at second, third, fourth.. call such as:

``` python
responses.add(responses.GET, 'http://twitter.com/api/1/foobar', json={"msg": "not found"}, status=404)
responses.add(responses.GET, 'http://twitter.com/api/1/foobar', json={"msg": "OK"}, status=200)
responses.add(responses.GET, 'http://twitter.com/api/1/foobar', json={"msg": "OK"}, status=200)
responses.add(responses.GET, 'http://twitter.com/api/1/foobar', json={"msg": "not found"}, status=404)
print requests.get('http://twitter.com/api/1/foobar')
print requests.get('http://twitter.com/api/1/foobar')
print requests.get('http://twitter.com/api/1/foobar')
print requests.get('http://twitter.com/api/1/foobar')
```

Expected standard output:

```
<Response [404]>
<Response [200]>
<Response [200]>
<Response [404]>
```

Explanation: If we add response to same URL we simply specify the next response.

Just like it is done in PHP using PHPUnit, provided with [that](https://phpunit.de/manual/current/en/test-doubles.html)

Refer to `Table 9.1. Matchers`, quoting 

screen shot 2016-01-27 at 4 52 21 pm

For not the only way I see to reproduce such behaviour is using callback with counter which is a global variable, and it doesn't seem to be elegant way to do it.

``` python
global call

@responses.activate
def test():
    global call
    call = 0
    def request_callback(request):
        global call
        headers = {'key': 'value'}
        if call == 0 :
            call = 1
            return (404, headers, '{"msg": "not found"}')
        elif call == 1 :
            call = 0
            return (200, headers, '{"msg": "ok"}')
    responses.add_callback(responses.GET, 'http://twitter.com/api/1/foobar', callback=request_callback)
    print requests.get('http://twitter.com/api/1/foobar')
    print requests.get('http://twitter.com/api/1/foobar')
```

### LINKED ISSUE #135: Test Max Retries

### LINKED ISSUE #376: Error or better document overlapping regex matcher behaviour

### LINKED ISSUE #406: Searching Responses and Manipulation of Matching Order
## Description

Currently, the order of registered responses is only determined by the insertion order which is also the order in which a request will look for matches. #400 wants to introduce a priority attribute which provides a little more control on the matching order. However, more control over the responses within a priority group in the registry would be beneficial. It should also be possible to locate (get matching order index) a response in the registry and search for responses of some criteria.

## User Stories

    - I want to be able to specify the matching order of newly registered responses at any time, so that I do not need 
      to insert my responses in metching order.
    - I want to be able to search for registered responses, so I can more easily update or remove them.
    - I want to be able to move registered responses (without changing priority), so I can manipulate the matching
      order at any time.

### Search interface

```python
class RequestsMock:
    ...
    def search(self, resp_filter: Callable[[response], bool]) -> list[BaseResponse]:
        ...
        return responses
    ...
resp = RequestMock()
for response in my_responses:
    resp.add(respons)
index, relevant = resp.search(lambda response: response.method == responses.POST)
```

### Change Matching Order

```python
class RequestsMock:
    ...
    def move(self, response: Union[BaseResponse, int], pos: int, relative: bool = False) -> None:
        ...
    ...
resp = RequestMock()
for response in my_responses:
    resp.add(response)
last_response = my_responses[-1]
# move three position forward
resp.move(last_response, -3, relative=True)
# move in front
resp.move(last_response, 0)
# move last response in front (access by index) this might be useful if the response is registerd 
# several times, because the non-index approach should only move the first occurring response.
resp.move(-1, 0)
```
Note that this method should try to move the response to the desired location, but it might be sorted backwards because of a lower priority.
Example (only showing priority).
In `[HIGH, HIGH, NORMAL, LOW]` using `move(-1, 0)` would not change the list, because the last element is the only one with priority low, but `move(1, 0)` would swap the first two items.

### Set Matching Position Directly

```python
class RequestsMock:
    ...
    def add(self, ..., pos: Optional[int] = None) -> ...:
        ...
    ...
resp = RequestMock()
for respons in my_responses:
    resp.add(respons)
# add this response on third position (use python indexing starting with 0)
resp.add(my_other_response, pos=2)
```
Also here, the position may be sorted backwards depending on the priority.

The methods `add` and `move` could also get a flag called `force` which would overwrite the priority to enforce the desired location, but I do not know if this is needed or should be the default.

Looking forward for some feedback to the proposed feature.

beliaev-maksim:
Added support for the rotating responses (dependent on invocation index)

Closes #52
Closes #99
Closes #135
Closes #376
Closes #406

I think all of the above mentioned Issues could be sorted out by this PR or by already implemented Registry mechanism


responses/test_registries.py
```diff
@@ -68,3 +71,91 @@ class CustomRegistry(registries.FirstMatchRegistry):
...
+                json={"msg": "not found"},
+                status=404,
+            )
+
+            resp = requests.get("http://twitter.com/api/1/foobar")
+            assert resp.status_code == 404
+            resp = requests.get("http://twitter.com/api/1/foobar")
+            assert resp.status_code == 200
+            resp = requests.get("http://twitter.com/api/1/foobar")
+            assert resp.status_code == 200
```
markstory, line 110: Should we also check the response bodies or use more unique status codes to ensure the ordering?

beliaev-maksim: Agree, let me set unique codes and bodies for each request-response pair

beliaev-maksim: @markstory 
done

responses/registries.py
```diff
@@ -67,3 +67,25 @@ def replace(self, response: "BaseResponse") -> None:
...
+        self, request: "PreparedRequest"
+    ) -> Tuple[Optional["BaseResponse"], List[str]]:
+
+        if not self.registered:
+            return None, ["No more registered responses"]
+
+        response = self.registered.pop(0)
+        match_result, reason = response.matches(request)
+        if not match_result:
+            self.reset()
```
markstory, line 83: Won't wiping the registry remove other mocks that haven't been called yet?

responses/registries.py
```diff
@@ -67,3 +67,25 @@ def replace(self, response: "BaseResponse") -> None:
...
+            self.add(response)
+            reason = (
+                "Next 'Response' in the order doesn't match "
+                f"due to the following reason: {reason}."
```
markstory, line 87: This is a pretty good error message. I think it could be better if we included what the next expected response URL is as well.

responses/registries.py
beliaev-maksim: @markstory 
yes, it will
the point is that if we want to modify error message without changing existing code, then we need to clean up all registered `responses` and fill only existing one

responses/registries.py
beliaev-maksim: @markstory 
we get something like in test 
```
"- GET http://twitter.com/api/1/barfoo Next 'Response' in the order doesn't match due to the following reason: URL does not match"
```

do you want to add at the end smth like: `Expected: <url>` ?

beliaev-maksim: now this looks like this, further changes in the message will require changes in `__init__.py` which I hesitate a bit.

However, I am fine to change wording in the `reason`

```
Connection refused by Responses - the call doesn't match any registered mock.

Request: 
- GET http://twitter.com/api/1/foobar

Available matches:
- GET http://twitter.com/api/1/barfoo Next 'Response' in the order doesn't match due to the following reason: URL does not match.
```

markstory: > do you want to add at the end smth like: Expected: <url> ?

I think that would be better. The current error message you have where only one response is listed as available is pretty good. I wonder if developers will be tripped up by some of their mocks missing in the output.

beliaev-maksim: @markstory 
the problem is that then we need to rewrite current messaging.
here https://github.com/getsentry/responses/blob/0d2c4f9543fcdf445c8a838ba1b756f007fac02d/responses/__init__.py#L727 we iterate over all registered and report the issue in format (URL - Reason)

if we would like to change it, then it may affect tests that rely on the wording

in this scenario we are flexible only on changing the reason but considering that `Available matches` phrase will be still present.
ideas on meaningful output we can send?

beliaev-maksim: generally, it should be somehow logical, that in `OrderedRegistry` we have only one available match since all other are irrelevant

markstory: I don't have strong conviction that the message will be unclear. How about we move forward with what we have today and wait for feedback that the messaging is unclear or confusing.

beliaev-maksim: @markstory 
sure,
maybe somebody will propose a nice alternative :)

PR #482, opened 2022-01-21
minor cleanup

beliaev-maksim:
clean up after py2.7


beliaev-maksim:
I see that cov dropped. Means we did not have a test for that line..

will add it

PR #486, opened 2022-01-28
remove response_callback in exception

beliaev-maksim:
there is no reason to execute `response_callback` on Exceptions. We do not return back `Response` object, like we do on line: https://github.com/getsentry/responses/blob/ee46f922e4271b583c5415ae29d77f563a66e751/responses/__init__.py#L760


CHANGES
```diff
@@ -3,6 +3,7 @@
 
 * Expose `get_registry()` method of `RequestsMock` object. Replaces internal `_get_registry()`.
 * Added support for `async/await` functions.
+* `response_callback` is not executed on failed `Response`s
```
markstory, line 6:
```suggestion
* `response_callback` is no longer executed on exceptions raised by failed `Response`s
```

PR #487, opened 2022-01-31
Fix duplicate content-type silently failing

### LINKED ISSUE #483: regression regarding content_type in 0.10.8
A test started failing on an application when updating our code and testing against the current `responses`.

I traced it back to a regression in the 0.10.8 release. When there is collision between `content_type` and `adding_headers["Content-Type"]` on `responses.RequestsMock.add`

In 0.10.7 and below, the `content_type` argument is ignored if `Content-Type` is provided as a header.

In 0.10.8 and above, the `content_type` argument is prepended to  the headers' `Content-Type` when provided.

### Steps to Reproduce

An executable reproduction is provided below.

pip install "responses==0.10.7"
python {ExampleFile}.py

pip install --upgrade responses
python {ExampleFile}.py

### Expected Result

I did not expect a breaking change.

I would expect a breaking change like this to either:

* be properly listed in the changelog
* raise an exception when there is an overlap in conflicting arguments

When trying to decide workarounds, this was also noted:

When headers are computed, a "," is used to prepend the `content_type` to the `adding_headers["Content-Type"]`. This means one can not strip out "text/html" from a header that has a charset in it, and rely on Responses to upgrade it:

This is invalid:
 
     `{'Content-Type': 'text/html, charset=utf-8'}`

As this is valid 

     `{'Content-Type': 'text/html; charset=utf-8'}`

While I'd love to see the old functionality return, I understand fixing this regression may not be a priority.  Because the Content-Type generated by passing in both arguments is not RFC compliant,  I think a changlog entry is insufficient and this library would need to raise an Exception if `content_type` is specified alongside a "Content-Type" header.  

This was painful to debug.

### Actual Result

Described above.

### Executable Example

```python
# pypi
import responses
import requests

with responses.RequestsMock(assert_all_requests_are_fired=False) as rsps:
    rsps.add(
        responses.GET,
        "http://example.com/index-1.html",
        body="<html>1</html>",
        content_type="text/html",
    )
    rsps.add(
        responses.GET,
        "http://example.com/index-2.html",
        body="<html>2</html>",
        adding_headers={
            "Content-Type": "text/html2; charset=utf-8",  # html2 to differentiate against html
        },
    )
    rsps.add(
        responses.GET,
        "http://example.com/index-3.html",
        body="<html>3</html>",
        content_type="text/html",
        adding_headers={
            "Content-Type": "text/html2; charset=utf-8",  # html2 to differentiate against html
        },
    )

    r1 = requests.get("http://example.com/index-1.html")
    assert r1.headers["Content-Type"] == "text/html"
    
    r2 = requests.get("http://example.com/index-2.html")
    try:
        assert r2.headers["Content-Type"] == "text/plain, text/html2; charset=utf-8"
        print("r2 REGRESSION")
    except:
        assert r2.headers["Content-Type"] == "text/html2; charset=utf-8"
        print("r2 legacy")

    r3 = requests.get("http://example.com/index-3.html")
    try:
        assert r3.headers["Content-Type"] == "text/html, text/html2; charset=utf-8"
        print("r3 REGRESSION")
    except:
        assert r3.headers["Content-Type"] == "text/html2; charset=utf-8"
        print("r3 legacy")
```

markstory:
When both the `content_type` keyword argument and the `Content-Type` header are defined we have an ambiguous set of parameters. Raise an error to let the developer known they have made an invalid method call.

Fixes #483

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -566,7 +566,7 @@ class RequestsMock(object):
         if adding_headers is not None:
             kwargs.setdefault("headers", adding_headers)
         if "content_type" in kwargs and "headers" in kwargs:
-            header_keys = list(map(lambda x: x.lower(), kwargs["headers"].keys()))
+            header_keys = [header.lower() for header in kwargs["headers"]]
             if "content-type" in header_keys:
                 raise RuntimeError(
                     "You cannot define both `content_type` and `headers[Content-Type]`."
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -565,6 +565,13 @@ class RequestsMock(object):
 
         if adding_headers is not None:
             kwargs.setdefault("headers", adding_headers)
+        if "content_type" in kwargs and "headers" in kwargs:
+            header_keys = list(map(lambda x: x.lower(), kwargs["headers"].keys()))
+            if "content-type" in header_keys:
+                raise RuntimeError(
+                    "You cannot define both `content_type` and `headers[Content-Type]`."
+                    " Using the `content_type` kwarg is recommended."
+                )
 
         self._registry.add(Response(method=method, url=url, body=body, **kwargs))
 
--- a/CHANGES
+++ b/CHANGES
@@ -3,6 +3,7 @@
 
 * Expose `get_registry()` method of `RequestsMock` object. Replaces internal `_get_registry()`.
 * Added support for `async/await` functions.
+* An error is now raised when both `content_type` and `headers[content-type]` are provided as parameters.
 
 0.18.0
 ------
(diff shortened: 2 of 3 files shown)


responses/__init__.py
```diff
@@ -565,6 +565,13 @@ def add(
 
         if adding_headers is not None:
             kwargs.setdefault("headers", adding_headers)
+        if "content_type" in kwargs and "headers" in kwargs:
+            header_keys = list(map(lambda x: x.lower(), kwargs["headers"].keys()))
```
beliaev-maksim, line 569: could this be simplified to?
```suggestion
            header_keys = [header.lower() for header in kwargs["headers"]]
```

PR #488, opened 2022-01-31
added isort pre-commit hook

beliaev-maksim:
* added `isort` pre-commit hook to have readable imports
* bump pre-commit version in Makefile

otherwise it is a mess in imports, I cannot find where and what we import :D

let's merge it after #462, #477, #487, #489

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -1,25 +1,24 @@
-from http import client
-
 import inspect
-from http import cookies
 import json as json_module
 import logging
+from collections import namedtuple
+from collections.abc import Sequence
+from collections.abc import Sized
+from functools import wraps
+from http import client
+from http import cookies
 from itertools import groupby
 from re import Pattern
+from warnings import warn
 
-
-from collections import namedtuple
-from functools import wraps
 from requests.adapters import HTTPAdapter
 from requests.exceptions import ConnectionError
 from requests.utils import cookiejar_from_dict
+
 from responses.matchers import json_params_matcher as _json_params_matcher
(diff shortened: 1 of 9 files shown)


responses/__init__.py
```diff
@@ -1,25 +1,24 @@
...
 import inspect
-from http import cookies
 import json as json_module
 import logging
+from collections import namedtuple
+from collections.abc import Sequence
+from collections.abc import Sized
+from functools import wraps
+from http import client
+from http import cookies
```
markstory, line 9: Shouldn't these be grouped?

responses/__init__.py
```diff
@@ -34,20 +33,15 @@
...
 from io import BufferedReader
-
+from io import BytesIO
 from unittest import mock as std_mock
+from urllib.parse import parse_qsl
+from urllib.parse import quote
+from urllib.parse import urlparse
+from urllib.parse import urlsplit
+from urllib.parse import urlunparse
+from urllib.parse import urlunsplit
```
markstory, line 44: Same here with grouped imports.

responses/__init__.py
beliaev-maksim: @markstory 
this is better in terms of visibility when you add/remove imports. GIT handles this as separate lines, so you easily see what was added per PR

for human readability it is also fine, since everything is sorted alphanumerically.

beliaev-maksim: here is another lib, that I would not use due to some issues that I faced during pipelines
but it has a pretty good explanation: https://github.com/asottile/reorder_python_imports#why-this-style

PR #494, opened 2022-02-08
mbeliaev/shield

beliaev-maksim:
add shield with downloads per month
image


beliaev-maksim:
tests fail, #495 should fix it

markstory:
Thank you 👍

PR #497, opened 2022-02-09
remove readonly properties

beliaev-maksim:
currently we expose 2 attributes, users may think that changing them will affect something, but it is not
best example to show:

```python
import responses

print(responses.assert_all_requests_are_fired)
responses.assert_all_requests_are_fired = True
print(responses.assert_all_requests_are_fired)
print(responses.mock.assert_all_requests_are_fired)
```

produces
```python
False
True
False
```


beliaev-maksim:
tests fail, #495 should fix it

markstory:
I'm concerned that we'll break user-land code by taking these away. Given the impacts of past clean up changes, we've seen that user-land code is using several internals (eg. `_is_string`). Perhaps we should add a property wrapper for these and emit deprecation warnings?

beliaev-maksim:
also a valid argument

any ideas how to create a property without a class instance?
since now it is just a variable in the module, eg `responses.assert_all_requests_are_fired`

aburgel:
> any ideas how to create a property without a class instance?

you can do something like this to deprecate these module attributes: https://www.python.org/dev/peps/pep-0562/

beliaev-maksim:
> > any ideas how to create a property without a class instance?
> 
> you can do something like this to deprecate these module attributes: https://www.python.org/dev/peps/pep-0562/

Sounds like an idea, I will try it
Thanks

beliaev-maksim:
@markstory 
can you please review?

beliaev-maksim:
@markstory 
merge conflicts are resolved

PR #498, opened 2022-02-09
mbeliaev/loose matcher

### LINKED ISSUE #496: Partial query_param_matcher
Currently it seems that the query_param_matcher, matcher, only performs a match when **ALL** query parameters are matched.

Is it possible to add an option to the function to allow for `loose` (probably a much better word for that) matching, where a match against any of the query parameters results in the tuple returning true?

beliaev-maksim:
Closes #496

### DIFF
--- a/responses/matchers.py
+++ b/responses/matchers.py
@@ -181,7 +181,9 @@ def query_param_matcher(
                 _create_key_val_str(params_dict),
             )
             if not strict_match:
-                reason += "\nNote that you use not strict parameters check."
+                reason += (
+                    "\nYou can use `strict_match=True` to do a strict parameters check."
+                )
 
         return valid, reason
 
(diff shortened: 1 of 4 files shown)


beliaev-maksim:
tests fail, https://github.com/getsentry/responses/pull/495 should fix it

responses/matchers.py
```diff
@@ -152,17 +166,22 @@ def match(request: PreparedRequest) -> Tuple[bool, str]:
...
+
+        valid = sorted(params_dict.items()) == sorted(request_params_dict.items())
 
         if not valid:
             reason = "Parameters do not match. {} doesn't match {}".format(
                 _create_key_val_str(request_params_dict),
                 _create_key_val_str(params_dict),
             )
+            if not strict_match:
+                reason += "\nNote that you use not strict parameters check."
```
markstory, line 184:
```suggestion
                reason += "\nYou can use `strict_match=False` to use enable matches."
```

What do you think of this? It makes the error more instructive on what a next step could be.

beliaev-maksim: agree with the idea, but I think the message should be `"You can use 'strict_match=True' to do a strict parameters check."`
since now this reason is triggered when `strict_match=False`

PR #500, opened 2022-02-10
mbeliaev/patcher

### LINKED ISSUE #481: Mocking is not completely undone when decorated tests are nested
First of all, I'd like to point out that I only noticed this _problem_ in a project where HTTP requests are mocked with multiple libraries (I know, bad style).

### Environment

* Ubuntu 20.04
* Python 3.X, X in (6, 8)
* responses 0.17.0
* httpretty 1.1.4

### Steps to Reproduce

1. As mentioned, I ended up with a testsuite that uses two libraries for mocking: `responses` and `httpretty`.
2. In an effort to reduce code duplication in the test suite, I wanted to re-use a test method decorated with `responses.activate` within another test method, that is also decorated with `responses.activate`.
3. Another test method decorated with `httpretty.activate` will fail, if it is executed after the test including the nesting is executed.

To illustrate this, I have created a [gist](https://gist.github.com/crazyscientist/d254f9d31bf57b97f277dde5f9e205af).

### Expected Result

Mocks introduced my the `responses` package are completely rolled back, even if there is a nesting of mocks.

### Actual Result

To illustrate this, I have created a [gist](https://gist.github.com/crazyscientist/d254f9d31bf57b97f277dde5f9e205af):

- In it's current state (no nesting of tests decorated with `responses.activate`), all tests in the gist will pass.
- If the decorator in line 29 is commented in, one test will fail:

```
======================================================================
ERROR: test_c_httpretty (scratch_4.TestRequest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/andi/virtualenvs/smash/lib/python3.8/site-packages/httpretty/core.py", line 2075, in wrapper
    return test(*args, **kw)
  File "/home/andi/.config/JetBrains/PyCharm2021.2/scratches/scratch_4.py", line 29, in test_c_httpretty
    self.assertEqual(b"Hello 3", requests.get("http://example.com/3").content)
  File "/home/andi/virtualenvs/smash/lib/python3.8/site-packages/requests/api.py", line 75, in get
    return request('get', url, params=params, **kwargs)
  File "/home/andi/virtualenvs/smash/lib/python3.8/site-packages/requests/api.py", line 61, in request
    return session.request(method=method, url=url, **kwargs)
  File "/home/andi/virtualenvs/smash/lib/python3.8/site-packages/requests/sessions.py", line 529, in request
    resp = self.send(prep, **send_kwargs)
  File "/home/andi/virtualenvs/smash/lib/python3.8/site-packages/requests/sessions.py", line 645, in send
    r = adapter.send(request, **kwargs)
  File "/home/andi/virtualenvs/smash/lib/python3.8/site-packages/responses/__init__.py", line 842, in unbound_on_send
    return self._on_request(adapter, request, *a, **kwargs)
  File "/home/andi/virtualenvs/smash/lib/python3.8/site-packages/responses/__init__.py", line 821, in _on_request
    raise response
requests.exceptions.ConnectionError: Connection refused by Responses - the call doesn't match any registered mock.

Request: 
- GET http://example.com/3

Available matches:

```

beliaev-maksim:
Closes #481

that was a tricky one, after research I found the root cause. Since default decorator always uses `mock = _default_mock`, when we apply double patching, we override the value of the `_patcher`. When standard python `mock` is trying to stop the patcher, it fails, since we have another object in the memory.

Keeping only single `mock` for all nested functions is sufficient enough. Thus, prevent from override and do clean up on the `__exit__`.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -4,7 +4,7 @@
 * Added a registry that provides more strict ordering based on the invocation index.
   See `responses.registries.OrderedRegistry`.
 * Prevent `responses.activate` decorator to leak, if wrapped function called from within another
-  wrapped function. Also, allow above mention chain calling. See #481 for more details.
+  wrapped function. Also, allow calling of above mentioned chain. See #481 for more details.
 * Expose `get_registry()` method of `RequestsMock` object. Replaces internal `_get_registry()`.
 * `query_param_matcher` can now accept dictionaries with `int` and `float` values.
 * Added support for `async/await` functions.


responses/__init__.py
```diff
@@ -809,11 +810,16 @@ def start(self):
         def unbound_on_send(adapter, request, *a, **kwargs):
             return self._on_request(adapter, request, *a, **kwargs)
 
-        self._patcher = std_mock.patch(target=self.target, new=unbound_on_send)
-        self._patcher.start()
+        if not self._patcher:
```
markstory, line 813: Should the function on line 810 only be defined when we're creating a patcher as well?

crazyscientist:
Wow, thanks for fixing this :partying_face:

PR #505, opened 2022-02-21
Add logging of passthrus when there's no match

aburgel:
A couple of changes related to passthrus:

`responses.passthru_prefixes` is always empty because it is a value type that is captured when the responses module is loaded. I've fixed this by making passthru_prefixes a mutable list. Of course, this will allow users to add things directly to the list. The downsides of this seem minimal, but I'm happy to fix this in another way if we think it's important to maintain immutability.

I also added some logging of passthrus when there is no match. I found this helpful for debugging.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -786,7 +786,7 @@ class RequestsMock(object):
             if self.passthru_prefixes:
                 error_msg += "Passthru prefixes:\n"
                 for p in self.passthru_prefixes:
-                    error_msg += "{}\n".format(p)
+                    error_msg += "- {}\n".format(p)
 
             response = ConnectionError(error_msg)
             response.request = request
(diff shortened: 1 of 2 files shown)


beliaev-maksim:
I am -1 on making prefixes mutable. 
We already had issue with leakage of prefixes

this issue is more related to the PR #497. I would add `responses.passthru_prefixes` to deprecation

please use `responses.mock.passthru_prefixes` instead

aburgel:
> I am -1 on making prefixes mutable.

I removed those changes. Now this PR just adds logging of passthrus when there's no match.

beliaev-maksim:
@aburgel
Did you follow contribution guide? You can easily set dev environment following it. 

Then, please, run tox and fix coverage

aburgel:
@beliaev-maksim I updated a test to get back to 100% coverage

PR #512, opened 2022-03-11
Removed internal `_cookies_from_headers` function

beliaev-maksim:
> Note: merge after #497 due to the potential merge conflict


markstory:
Going to hold off on this until 0.21.0 as 0.19 was a bit bumpy and I'd like to get the fixes out for folks using moto.

beliaev-maksim:
@bblommers 
just to double check, don't you rely on this functionality?

bblommers:
No, we don't use this particular field @beliaev-maksim - thanks for asking!

PR #517, opened 2022-03-17
mbeliaev/docs depr

beliaev-maksim:
here is look and feel
image

@markstory 
do you think we need to add info about removed internals as well?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,38 @@ Installing
 ``pip install responses``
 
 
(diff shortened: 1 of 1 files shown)


markstory:
>  do you think we need to add info about removed internals as well?

I don't think so. Users *shouldn't* have been using them and I don't think we should be on the hook for them using features that weren't meant to be used.

PR #520, opened 2022-03-17
add bug report template

### LINKED ISSUE #519: new template for `responses` issues
@chadwhitacre @markstory 
can we set a template for `responses` issues?
currently it is more about `sentry`

we need in template:
1. version of responses
2. minimal reproducible self-contained snippet. Ready to be copy pasted and run.
3. current output
4. expected output

now it looks like:
image

beliaev-maksim:
closes #519 

@markstory @chadwhitacre 
you will need to configure template from the project settings (I do not have permissions to edit this)


chadwhitacre:
This is okay, but if we use the newer yaml format we will get a nice form vs. a textarea. Here's an example:

https://github.com/getsentry/sentry/blob/master/.github/ISSUE_TEMPLATE/bug.yml

raw: https://raw.githubusercontent.com/getsentry/sentry/master/.github/ISSUE_TEMPLATE/bug.yml

result: https://github.com/getsentry/sentry/issues/new?assignees=&labels=&template=bug.yml

What do you think? :)

beliaev-maksim:
@chadwhitacre 
As far as I know this is a beta feature. If you have already tested it and adopted, then we can switch here as well

beliaev-maksim:
@chadwhitacre @markstory 
here is look and feel of the form: https://github.com/beliaev-maksim/responses/blob/mbeliaev/template/.github/ISSUE_TEMPLATE/bug_report.yml

chadwhitacre (approved):
wfm

You good with this, @markstory?

markstory:
Looks great!

PR #522, opened 2022-03-21
mbeliaev/mypy

beliaev-maksim:
move some types to inlines


beliaev-maksim:
@markstory 
can you please check this PR? should be an easy one

I will resolve merge conflicts on #524 later

beliaev-maksim:
@markstory
Thanks!

PR #523, opened 2022-03-22
add redirections example and test

### LINKED ISSUE #152: future feature request - ability to mock exceptions with a response
I don't think this is possible in the current library, so for the future, it would be nice to be able to mock exceptions that have a response attribute.

As an example...

We have a flow of redirects:

    A -> 301 redirect -> B
    B -> 301 redirect -> C
    C -> 301 redirect -> D
    D -> 200 html

And I need to mock a ConnectionError, or otherwise, on C.

In a real-world situations, the error raised by `requests` has a `response` attribute, which will have the `.url` and `.history` attributes which can be inspected.  This information is unfortunately gone under `responses`.

I don't see this as being possible in the current design. It's probably not possible at all.  I did want to note this as a wishlist item if anyone ever has an idea.

beliaev-maksim:
closes #152 

rendered RST: https://github.com/beliaev-maksim/responses/tree/mbeliaev/redirects#url-redirection

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -864,6 +864,69 @@ You can also add multiple responses for the same url:
         assert resp.status_code == 200
 
 
+URL Redirection
+---------------
+
+In the following example you can see how to create a redirection chain and add custom exception that will be raised
+in the execution chain and contain the history of redirects.
+
+..  code-block::
+
+    A -> 301 redirect -> B
+    B -> 301 redirect -> C
+    C -> connection issue
+
+..  code-block:: python
+
+    import pytest
+    import requests
+
+    import responses
+
+
+    @responses.activate
+    def test_redirect():
+        # create multiple Response objects where first two contain redirect headers
+        rsp1 = responses.Response(
+            responses.GET,
+            "http://example.com/1",
+            status=301,
+            headers={"Location": "http://example.com/2"},
+        )
+        rsp2 = responses.Response(
+            responses.GET,
+            "http://example.com/2",
+            status=301,
+            headers={"Location": "http://example.com/3"},
+        )
+        rsp3 = responses.Response(responses.GET, "http://example.com/3", status=200)
+
+        # register above generated Responses in ``response`` module
+        responses.add(rsp1)
+        responses.add(rsp2)
+        responses.add(rsp3)
+
+        # do the first request in order to generate genuine ``requests`` response
+        # this object will contain genuine attributes of the response, like ``history``
+        rsp = requests.get("http://example.com/1")
+        responses.calls.reset()
+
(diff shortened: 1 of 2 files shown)


beliaev-maksim:
@kbakk
can you please kindle review and share your thoughts?

README.rst
```diff
@@ -864,6 +864,68 @@ You can also add multiple responses for the same url:
...
+        rsp3 = responses.Response(responses.GET, "http://example.com/3", status=200)
+
+        # register above generated Responses in ``response`` module
+        responses.add(rsp1)
+        responses.add(rsp2)
+        responses.add(rsp3)
+
+        # do the first request in order to generate genuine ``requests`` response
+        # this object will contain genuine attributes of the response, like ``history``
+        rsp = requests.get("http://example.com/1")
```
kbakk, line 911: Suggest to reset the call list, else you'll get calls from both here and in the actual test (relevant only if you do asserts on `responses.calls` but that's quite common in the other sections). 

```suggestion
        rsp = requests.get("http://example.com/1")
        responses.calls.reset()
```

kbakk (commented):
Looks good - one small suggestion only.

README.rst
beliaev-maksim: @kbakk 
that makes sense if want to have a clean list
thank you for the review

responses/tests/test_responses.py
```diff
@@ -2129,3 +2129,49 @@ def run():
...
+            status=301,
+            headers={"Location": "http://example.com/2"},
+        )
+        rsp2 = responses.Response(
+            responses.GET,
+            "http://example.com/2",
+            status=301,
+            headers={"Location": "http://example.com/3"},
+        )
+        rsp3 = responses.Response(responses.GET, "http://example.com/3", status=200)
```
markstory, line 2150: Should we have a method for generating `ConnectionError` style responses that doesn't require getting the response value out and modifying it? Perhaps something like

```python
rsp = responses.Response(
  responses.GET,
  'http://example.com/3',
  body=requests.ConnectionError('something bad')
)
responses.add(rsp)
```

beliaev-maksim: it will be hard to achieve. With provided example we will raise an error and the object will not contain history.

thus, we first need http response (not our response) object and then we attach this object to the Exception. 
If we want to generate genuine Response, then we have to rewrite lots of things and duplicate some code from `requests` and probably apply our mock on a deeper (or maybe higher) level

markstory: This is probably an uncommon scenario that we don't need to design an ergonomic API for. If it comes up again we can figure out a higher-level API.

PR #529, opened 2022-03-29
mbeliaev/assert call count

### LINKED ISSUE #335: `assert_call_count` should be able to check urls without matching querystring

beliaev-maksim:
closes #335

Considering evolved functionality of `matchers` I do not see it valuable to continue expanding `assert_call_count`. I believe more preferable will be to inspect `Response` objects, since they include matchers and all further functionality

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -575,7 +575,7 @@ class RequestsMock(object):
         """
         if isinstance(method, BaseResponse):
             self._registry.add(method)
-            return
+            return method
 
         if adding_headers is not None:
             kwargs.setdefault("headers", adding_headers)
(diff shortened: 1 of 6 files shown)


responses/__init__.py
```diff
@@ -587,28 +586,29 @@ def add(
                     " Using the `content_type` kwarg is recommended."
                 )
 
-        self._registry.add(Response(method=method, url=url, body=body, **kwargs))
+        response = Response(method=method, url=url, body=body, **kwargs)
+        return self._registry.add(response)
 
     def delete(self, *args, **kwargs):
-        self.add(DELETE, *args, **kwargs)
+        return self.add(DELETE, *args, **kwargs)
```
markstory, line 593: Should we be updating the typehint/signature on these methods to include the new return value?

PR #530, opened 2022-03-29
Add  `threading.Lock()` 

### LINKED ISSUE #515: Flaky test in multithreading environment
### Environment

python 3.10
responses 0.19.0
requests 2.27.1
pytest 7.1.0

### Steps to Reproduce

```python
import threading
import requests
import responses
def test_example():
    n_threads = 4
    n_requests = 100
    with responses.RequestsMock() as m:
        for i in range(n_requests):
            for j in range(n_threads):
                m.add(url=f"http://example.com/example{i%5}", method="GET")
        def fun():
            for i in range(n_requests):
                requests.get(f"http://example.com/example{i%5}")
        threads = [threading.Thread(name=f"example{i}", target=fun) for i in range(n_threads)]
        for thread in threads:
            thread.start()
        for thread in threads:
            thread.join()
```

Run this test with pytest.

### Expected Result

This test should always pass.

### Actual Result

This test sometimes doesn't pass, to increase frequency add more threads or run pytest with option --cov

This looks like a race condition in matching requests, adding a lock in the method `_find_match` seems to fix the issue.

beliaev-maksim:
closes #515 

based on the example in #515:
the issue why the error is raised, because `assert_all_requests_are_fired` checks for `call_count` on `Response` object.

in threading we may access `Response` in parallel, thus, `call_count > 1`. While users expects it to be `call_count == 1`.
Since we cannot ensure this, we need to Lock the thread

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 0.21.0
 ------
 
+* Add `threading.Lock()` to allow `responses` working with `threading` module.
 * Removed internal `_cookies_from_headers` function
 
 0.20.0
--- a/README.rst
+++ b/README.rst
@@ -517,6 +517,11 @@ you can see, that status code will depend on the invocation order.
 
 .. code-block:: python
 
+    import requests
+
+    import responses
+    from responses.registries import OrderedRegistry
+
     @responses.activate(registry=OrderedRegistry)
     def test_invocation_index():
         responses.add(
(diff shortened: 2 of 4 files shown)


responses/registries.py
```diff
@@ -88,3 +88,26 @@ def find(
             return None, [reason]
 
         return response, []
+
+
+class LockRegistry(FirstMatchRegistry):
```
markstory, line 93: Should this be something that the basic/default registry does? Having responses 'just work' by default would be nice and the overhead of locking a thread in environments without contention should be low.

beliaev-maksim: @markstory 
my only concern is the overhead since it looks like a rare case.

but if we are sure that this is negligible, then I can refactor

PR #531, opened 2022-03-31
added documentation and tests how to use requestmock as mock.patch

### LINKED ISSUE #31: Allow applying @responses.activate to a class
Right now I have to apply the decorator to every method I want to mock responses in; I'd prefer to apply the decorator only to the `TestCase` declaration as my `TestCase` does http in every test method. This would keep requests consistent with the mock.patch behaviour.

E.g.:

```
@responses.activate
class CommandHandlerTestCase(unittest.TestCase):

    def test_function_1(self):
        responses.add(...)

    def test_function_2(self):
        responses.add(...)

    def test_function_3(self):
        responses.add(...)
```

beliaev-maksim:
closes #31

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -732,8 +732,11 @@ Responses as a context manager
         resp = requests.get('http://twitter.com/api/1/foobar')
         resp.status_code == 404
 
-Responses as a pytest fixture
------------------------------
+Using ``responses`` in unit tests
+---------------------------------
+
+Responses as a ``pytest`` fixture
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 .. code-block:: python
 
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -732,8 +732,11 @@ Responses as a context manager
         resp = requests.get('http://twitter.com/api/1/foobar')
         resp.status_code == 404
 
-Responses as a pytest fixture
------------------------------
+Using ``responses`` in unit tests
+---------------------------------
```
markstory, line 736: Is this additional heading necessary? Responses is really only used in unit tests.

beliaev-maksim: @markstory 
we need some title to group. May be you can suggest something?

core idea is that this chapter describes integration with `pytest` and `unittest`

beliaev-maksim: @markstory 
maybe `integration with unit test frameworks` ?

beliaev-maksim: I did a rename

PR #534, opened 2022-04-02
Added option to match blank values

### LINKED ISSUE #533: urlencoded matcher does not check empty dict values
### Describe the bug

I am asserting a request having blank value in a dict key but it is failing as it is omitted out by urlencoded matcher.

### Additional context

_No response_

### Version of `responses`

0.20.0

### Steps to Reproduce

```python
import responses

@responses.activate
def test_it():
    url = "http://httpbin.org/post"
    responses.add(responses.POST, url, match=[
        matchers.urlencoded_params_matcher({"a": "", "b": "abc"})
    ], json={"c": "d"})
    requests.post(url, body={"a": "", "b": "abc"})

```

### Expected Result

The test should pass

### Actual Result

The test fails with "Request body does not match" error

codeasashu:
Fixes #533


codeasashu:
As further improvement, the PR can accomodate `kwargs` for `parse_qsl` so that the dict can be formed as per the test suite of users. This will allow micro customizations of the request body dict.

responses/matchers.py
```diff
@@ -53,7 +53,9 @@ def list_to_str(input_list: List[str]) -> str:
     return key_val_str
 
 
-def urlencoded_params_matcher(params: Optional[Dict[str, str]]) -> Callable[..., Any]:
+def urlencoded_params_matcher(
+    params: Optional[Dict[str, str]], allow_blank: bool = False
```
beliaev-maksim, line 57:
```suggestion
    params: Optional[Dict[str, str]], *, allow_blank: bool = False
```

Need to force only kwargs

beliaev-maksim (changes requested):
Also, please update CHANGES file

All the rest looks good, thanks

responses/matchers.py
codeasashu: Fixed in [c50cdf7](https://github.com/getsentry/responses/pull/534/commits/c50cdf7e25597e538b2b887df38ab8eef2439dfc)

CHANGES
```diff
@@ -1,3 +1,8 @@
+0.21.1
```
beliaev-maksim, line 1: Please put it under 0.21
It is not yet released

CHANGES
```diff
@@ -1,3 +1,8 @@
...
+------
+
+* Added `allow_blank` param to urlencoded_params_matcher
```
beliaev-maksim, line 4:
```suggestion
* Added null value support in `urlencoded_params_matcher` via `allow_blank` keyword argument 
```

beliaev-maksim (changes requested):

CHANGES
codeasashu: done. I've rebased

beliaev-maksim (approved):

PR #537, opened 2022-04-08
mbeliaev/req types

### LINKED ISSUE #552: `py.typed` file not installed by default in setup.cfg
### Describe the bug

mypy returns the following error when type checking our project.

`error: Skipping analyzing "responses": module is installed, but missing library stubs or py.typed marker  [import]`

### Additional context

Using mypy version 0.942 and the master version of responses.

### Version of `responses`

commit 04c9e1545d3413e42123c89a9c9d52f86c7b4a36

### Steps to Reproduce

Run mypy on a project that includes `responses`

```python
import responses

print(responses)
```

### Expected Result

There should be no mypy errors.

### Actual Result

`error: Skipping analyzing "responses": module is installed, but missing library stubs or py.typed marker  [import]`

beliaev-maksim:
finalize inline types migration. Remove stub file and go with python 3 inline style

closes #552


beliaev-maksim:
@markstory 
this was a real hell, still have to keep couple of type ignores.
I think we can handle ignores in another PR, if get a good idea how to fix them, for the moment I am lacking of ideas

also, can we ensure that #469 actually works and we do not get regression
I do not think we did any additional testing to check it

responses/__init__.py
```diff
@@ -756,11 +781,20 @@ def remove(self, method_or_response=None, url=None):
...
+            assert isinstance(method_or_response, str)
             response = BaseResponse(method=method_or_response, url=url)
 
         return self._registry.remove(response)
 
-    def replace(self, method_or_response=None, url=None, body="", *args, **kwargs):
+    def replace(
+        self,
+        method_or_response: Optional[Union[str, BaseResponse]] = None,
+        url: "Optional[Union[Pattern[str], str]]" = None,
```
markstory, line 793: Should we have type aliases for this and `Optional[Union[str, BaseResponse]]`? They are both used a few times.

beliaev-maksim: good idea, I added them

PR #544, opened 2022-04-14
remove setup.cfg

beliaev-maksim:
rationale:
* universal wheel is disabled by default
* yapf is not used in this project
* flake8 is running as a part of `pre-commit`, see #525
* pytest config is optional, and does not harm to have an `auto` traceback


markstory:
Thank you 🎉

PR #545, opened 2022-04-14
mbeliaev/recorder

beliaev-maksim:
propose beta feature for 0.22.0 to record responses to files using `man in the middle`

there are still steps to polish within this PR (types, docs, style, etc), but before doing this I need to confirm the idea.
General workflow looks like:
```python
import requests
from responses import _recorder


def another():
    rsp = requests.get("https://httpstat.us/500")
    rsp = requests.get("https://httpstat.us/202")

@_recorder.record(file_path="out.yml")
def test_recorder():
    rsp = requests.get("https://httpstat.us/404")
    rsp = requests.get("https://httpstat.us/201")
    another()
```

which will product following `out.yml` file:
```yaml
responses:
- response:
    body: ''
    content_type: text/plain
    headers: null
    method: GET
    status: 404
    url: https://httpstat.us/404
- response:
    body: ''
    content_type: text/plain
    headers: null
    method: GET
    status: 201
    url: https://httpstat.us/201
- response:
    body: ''
    content_type: text/plain
    headers: null
    method: GET
    status: 500
    url: https://httpstat.us/500
- response:
    body: ''
    content_type: text/plain
    headers: null
    method: GET
    status: 202
    url: https://httpstat.us/202

```

later I am going to extend the functionality and add possibility to load these yaml files


responses/registries.py
```diff
@@ -73,11 +73,31 @@ def replace(self, response: "BaseResponse") -> "BaseResponse":
...
+                        "url": rsp.url,
+                        "body": rsp.body,
+                        "status": rsp.status,
+                        "headers": rsp.headers,
+                        "content_type": rsp.content_type,
+                        "auto_calculate_content_length": rsp.auto_calculate_content_length,
+                    }
+                }
+            )
+        yaml.dump(data, destination)
```
markstory, line 94: I'm not sure yaml is the right format to use for this given how ambiguous YAML syntax is and how awkward YAML is at defining multi-line strings. TOML might be a less troublesome default, or we leave serialization undefined and users can supply their own 'save/load' implementation.

responses/_recorder.py
```diff
@@ -0,0 +1,98 @@
...
+from unittest import mock as std_mock
+from urllib.parse import parse_qsl
+from urllib.parse import urlsplit
+
+from responses import Response
+from responses import _real_send
+from responses.registries import OrderedRegistry
+
+
+class Recorder:
```
markstory, line 12: There is a good amount of duplication between the existing responses activation code and this. Should we first focus on loading 'recordings' and not on creating the 'man in the middle' to generate recordings?

I see the recording loader as the more valuable piece as it lets you refactor existing tests to share setup code or response fixtures.

responses/registries.py
beliaev-maksim: I thought about JSON, but with JSON there is a limitation of comments. I could expect that some users may want to add some notes to the recorded files

responses/_recorder.py
beliaev-maksim: that is the problem. I think first we need to decide how to write the data, since this might be more tricky. Then we can read it back

since if we go in opposite direction, we may create an incompatible interface


Activation and some parts could be refactored, for sure.
Now we need to validate the idea and the best output format

responses/registries.py
beliaev-maksim: my usage of toml is limited to `pyproject.toml` only. Do you think this will be more powerful format?

I see some good signs for it, since in py3.10 CPy has built-in toml parser, if I remember correctly

responses/_recorder.py
markstory: Isn't the data format dictated by what we need to load though? If we can't "record" all of that data we could record as much as possible. One wrinkle I see with recording is how custom registries work. I don't think we want to persist the registry state into the response file.

Would a minimal solution be to focus on loading a list of response definitions (without worrying about function responses) and then call `responses.add()` for each item in the bundle of responses?

beliaev-maksim:
@markstory 
I removed redundant code by inheriting from `RequestMock`
also, switched to `toml` format and added type hints in the code

if we are fine with the idea of implementation, then I will add tests to increase coverage

so, following code 

```python
import requests
from responses import _recorder


def another():
    rsp = requests.get("https://httpstat.us/500")
    rsp = requests.get("https://httpstat.us/202")

@_recorder.record(file_path="out.toml")
def test_recorder():
    rsp = requests.get("https://httpstat.us/404")
    rsp = requests.get("https://httpbin.org/status/wrong")
    another()
```
will product next output in out.toml
```toml
[[responses]]

[responses.response]
method = "GET"
url = "https://httpstat.us/404"
body = "404 Not Found"
status = 404
content_type = "text/plain"
auto_calculate_content_length = false
[[responses]]

[responses.response]
method = "GET"
url = "https://httpbin.org/status/wrong"
body = "Invalid status code"
status = 400
content_type = "text/plain"
auto_calculate_content_length = false
[[responses]]

[responses.response]
method = "GET"
url = "https://httpstat.us/500"
body = "500 Internal Server Error"
status = 500
content_type = "text/plain"
auto_calculate_content_length = false
[[responses]]

[responses.response]
method = "GET"
url = "https://httpstat.us/202"
body = "202 Accepted"
status = 202
content_type = "text/plain"
auto_calculate_content_length = false
```

responses/registries.py
```diff
@@ -73,6 +79,31 @@ def replace(self, response: "BaseResponse") -> "BaseResponse":
         self.registered[index] = response
         return response
 
+    def _dump(self, destination: "io.IOBase") -> None:
```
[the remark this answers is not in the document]
beliaev-maksim: I decided to create it in the registry because this will allow to dump existing tests to the file
Eg
If user ran `responses.add()` multiple times, then user can run `responses.get_registry()._dump(*args)`

That will allow safe (compared to brand new recording) transfer to the new method if required

In the future we can provide `serializer` argument, but at the moment I am a bit reluctant to support other serializers than one we decide

Also, user can also overwrite this method after inheriting

Or what are the other options to introduce this method?
As a separate function?

beliaev-maksim:
@markstory 
can you please review?

responses/registries.py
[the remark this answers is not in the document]
markstory: Couldn't dumping be on the `Recorder()`?  I'm concerned that recording feels like it is half in the Recorder and half in the registry right now.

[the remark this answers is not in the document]
beliaev-maksim: @markstory 
done!

this could be done even as a separate function. Then it will be possible to easily dump existing registries into the file

beliaev-maksim:
@markstory
Can you check pls?

Issue #546, opened 2022-04-17
Add ability to get a call by method and URL


paulnicolet:
When mocking a lot of requests, it might not always be desirable to access them by index

* We might not be interested in the calls order
* Adding new calls will shift all indices in assertions

```
responses.add(responses.GET, "http://example.com", body=b"test")
requests.get("http://example.com")

# The two are equivalent
responses.calls[0]
responses.calls.get(responses.GET, "http://example.com")
```

beliaev-maksim:
hi @paulnicolet 

have you tried a solution with a matcher ?
did it work for you?

PR #547, opened 2022-04-17
Get a call by method ans URL

paulnicolet:
Solves https://github.com/getsentry/responses/issues/546

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -217,6 +217,20 @@ class CallList(Sequence, Sized):
     def reset(self) -> None:
         self._calls = []
 
+    def get(self, method: str, url: str, idx: int = 0) -> Optional[Call]:
+        """Get a call by method and url. If multiple calls match, take the `idx` one."""
+        url = _ensure_url_default_path(url)
+        matches = [
+            call
+            for call in self
+            if call.request.method == method and _url_matches(call.request.url, url)
+        ]
+
+        if not matches:
+            return None
+
+        return matches[idx]
+
 
 def _ensure_url_default_path(
     url: "Union[Pattern[str], str]",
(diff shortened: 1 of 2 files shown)


beliaev-maksim:
I am not convinced that we need this extension. 

What is the exact issue you try to solve?

paulnicolet:
Sûre, I tried to describe in the linked issue, but the idea is that accessing calls by index makes assertions on requests pretty hard to read and maintain IMO, especially when having a good number of those. 

- Using indices forces to think about calls order, while we might not care
- It's hard for the reader to build understand which call are we accessing with a simple index
- Adding new calls will make all the indices shift, which is a pain to maintain

The idea is to have an easy way to fetch a given call, that exposes clearly the intention.

beliaev-maksim:
What is about using `matchers` to match request body?
https://github.com/getsentry/responses#matching-request-body-contents

paulnicolet:
Indeed. I have two concerns with matchers:

- It's less explicit than assertions to express that we are testing the body
- I'm not sure it's possible to extract custom fixtures out of requests mocks. It's a pattern that can come handy with POST requests, to define global fixtures and use those fixtures in a bunch of tests across the codebase. Then, we want to explicitly test that this call has been made with a precise body depending on the context, which we cannot do using matchers.

beliaev-maksim:
I still believe that this is achievable with `matchers`

Ideally, do a check by calls is less recommended compared to check by responses


I have some concerns with proposed implementation from this PR. Very often users use `re` to match (then it should be sorted here as well) and, also, returning calls by index in case of multiple matches returns us back to the same issue not giving any benefit 


If we can resolve the issue with existing functionality in a simple way, then I don't see a need to modify the code base

paulnicolet:
Of course, my proposal arises from the fact that I was not completely happy with the existing tools. 

Fine, let me put some extra effort and see if I can solve my issues using matchers, if I cannot I'll come back later and try to convince harder, with a more robust PR if needed 😉

PR #553, opened 2022-05-03
fix(ci) Fix lint errors from additional ignores

markstory:


beliaev-maksim:
that is so weird

I have to add these ignores since it was failing without them, see that all the checks passed in CI. 

Now it passes again without comments
why is that?

markstory:
> why is that?

No idea. Is it possible that mypy did a release?

beliaev-maksim:
> > why is that?
> 
> No idea. Is it possible that mypy did a release?

that was my first guess. But both current PR and mine use 0.950

PR #558, opened 2022-05-31
Allow configuration of passthrough during BaseResponse construction

### LINKED ISSUE #557: Feature request: allow configuration of passthrough during BaseResponse construction
### Use case

I'm working on a Python test suite that returns mock data for HTTP GET requests, and it has an 'online' mode during which no mocking of requests occurs.

I'd like to migrate the test suite to use `responses`, and would like to retain the online test functionality while also allowing for the assert-on-all-requests logic and other functionality.

This is already achievable using the [`passthrough`](https://github.com/getsentry/responses/tree/3d83a88382bf5cedefbbab3986bbfeeb07883fac#passing-through-real-requests) functionality -- but it's _not_ currently possible to provide `passthrough` as a keyword argument to `responses.add` nor the [shortcut methods](https://github.com/getsentry/responses/tree/3d83a88382bf5cedefbbab3986bbfeeb07883fac#shortcuts).

### Illustration

Providing that keyword argument on `BaseResponse` would allow simplification of code that looks like this:

```py
    for url, content in mock_requests:
        rsp = responses.Response(responses.GET, url, body=content)  # note: temporary local variable declared to hold a response
        rsp.passthrough = running_online
        rsps.add(rsp)
```

...to code that looks like this:

```py
    for url, content in mock_requests:
        rsps.get(url=url, body=content, passthrough=running_online)
```

Not a huge readability saving, perhaps - although I think it may be relatively straightforward to support, and I haven't yet thought of any downsides (but perhaps it isn't, for some intentional reason? - constructor argument pollution?).  I'll offer a pull request with a potential implementation.

### Related

Extends #399.

jayaddison:
Resolves #557.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,8 @@
+0.22.0
+------
+
+* Add support for the `passthrough` argument to the response shortcut methods
+
 0.21.0
 ------
 
--- a/CHANGES
+++ b/CHANGES
@@ -1,7 +1,7 @@
 0.22.0
 ------
 
-* Add support for the `passthrough` argument to the response shortcut methods
+* Add `passthrough` argument to `BaseResponse` object. See #557
 
 0.21.0
 ------
--- a/README.rst
+++ b/README.rst
@@ -1172,8 +1172,8 @@ need to allow an entire domain or path subtree to send requests:
     responses.add_passthru(re.compile("https://percy.io/\\w+"))
 
 
-Lastly, you can use the ``passthrough`` argument to the ``BaseResponse`` constructor
-to enable a response to behave as a pass through.
+Lastly, you can use the ``passthrough`` argument of the ``Response`` object
+to force a response to behave as a pass through.
 
 .. code-block:: python
 
--- a/README.rst
+++ b/README.rst
@@ -1172,14 +1172,18 @@ need to allow an entire domain or path subtree to send requests:
     responses.add_passthru(re.compile("https://percy.io/\\w+"))
 
 
-Lastly, you can use the ``response.passthrough`` attribute on ``BaseResponse`` or
-use ``PassthroughResponse`` to enable a response to behave as a pass through.
+Lastly, you can use the ``passthrough`` argument to the ``BaseResponse`` constructor
+to enable a response to behave as a pass through.
 
 .. code-block:: python
 
     # Enable passthrough for a single response
-    response = Response(responses.GET, "http://example.com", body="not used")
-    response.passthrough = True
+    response = Response(
+        responses.GET,
+        "http://example.com",
+        body="not used",
+        passthrough=True,
+    )
     responses.add(response)
 
     # Use PassthroughResponse
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -599,7 +599,7 @@ class CallbackResponse(BaseResponse):
 
 
 class PassthroughResponse(BaseResponse):
-    def __init__(self, *args, **kwargs):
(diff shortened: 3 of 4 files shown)


jayaddison:
> also, do we need then class attribute ?

Good question.  I've left it as-is for now for backwards-compatibility, although I have replaced the mention of it in the documentation.

Perhaps it could make sense to remove it.  Pattern-wise, perhaps we'd want to recommend that `Response` objects should be mostly/completely immutable after they are created?

README.rst
```diff
@@ -1172,14 +1172,18 @@ need to allow an entire domain or path subtree to send requests:
     responses.add_passthru(re.compile("https://percy.io/\\w+"))
 
 
-Lastly, you can use the ``response.passthrough`` attribute on ``BaseResponse`` or
-use ``PassthroughResponse`` to enable a response to behave as a pass through.
+Lastly, you can use the ``passthrough`` argument to the ``BaseResponse`` constructor
+to enable a response to behave as a pass through.
```
beliaev-maksim, line 1176:
```suggestion
Lastly, you can use the ``passthrough`` argument of the ``Response`` object
to force a response to behave as a pass through.
```

CHANGES
```diff
@@ -1,3 +1,8 @@
+0.22.0
+------
+
+* Add support for the `passthrough` argument to the response shortcut methods
```
beliaev-maksim, line 4:
```suggestion
* Add `passthrough` argument to `BaseResponse` object (#557)
```

beliaev-maksim (changes requested):

beliaev-maksim (changes requested):

responses/__init__.py
```diff
@@ -366,6 +368,8 @@ def __init__(
 
         self.match: "_MatcherIterable" = match
         self.call_count: int = 0
+        if "passthrough" in kwargs:
+            self.passthrough = kwargs["passthrough"]
```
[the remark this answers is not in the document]
jayaddison: Ah, my mistake - I tried that, but then saw some unit test failures.  And now I've realized that the tests were failing since the start.  Taking a look at this again now.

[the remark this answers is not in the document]
jayaddison: I'm not 100% certain why yet, but the change to use a named keyword argument caused problems for call sites using the `PassthroughResponse` directly (resolved by [adding an explicit super-constructor call](https://github.com/getsentry/responses/pull/558/commits/64f32f5c3b853a271cb61889e1b566727bae3edf#diff-27a24757cfb28e26971f5b4f49ba6f533b5a51c770d57369cebbef09c7433d6cR602-R609))

responses/__init__.py
```diff
@@ -600,7 +599,14 @@ def get_response(self, request: "PreparedRequest") -> HTTPResponse:
...
 class PassthroughResponse(BaseResponse):
-    passthrough: bool = True
+    def __init__(
+        self,
+        method: str,
+        url: _URLPatternType,
+        match_querystring: Union[bool, object] = None,
+        match: "_MatcherIterable" = (),
+    ):
+        super().__init__(method, url, match_querystring, match, passthrough=True)
```
beliaev-maksim, line 609:
```suggestion
    def __init__(self, *args, **kwargs):
        super().__init__(*args, passthrough=True, **kwargs)
```

markstory:
Thank you 🎉

jayaddison:
Thanks @markstory!

PR #568, opened 2022-07-15
build(setup.py): Add `project_urls` for PyPI

tony:
[`project_urls`](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#project-urls) for _setup.py_

For the [PyPI page](https://pypi.python.org/pypi/responses/) 

image


markstory:
Thank you 🎉

PR #569, opened 2022-07-28
Json matcher

### LINKED ISSUE #559: json_params_matcher does not support general JSONs (e.g. lists of dicts)
### Describe the bug

The implementation of `json_params_matcher` does not support arbitrarily nested JSON objects, such as lists of dicts. 

It says so in the signature: 

`def json_params_matcher(params: Optional[Dict[str, Any]]) -> Callable[..., Any]:`

and the simple check of `params_dict == json_body` is not enough. Additionally, an Exception is raised in this section, as a dict is expected.

`
            if not valid:
                reason = "request.body doesn't match: {} doesn't match {}".format(
                    _create_key_val_str(json_body), _create_key_val_str(params_dict)
                )
`

### Additional context

I can provide a PR, but there are different ways to tackle this:

- Custom function to recursively check => no dependencies
- Adopting something like https://github.com/seperman/deepdiff or https://github.com/xlwings/jsondiff to support with the extraction of reasons for JSONs being not equal

As this might entail additional dependencies, I'd like to get your opinion on this before submitting a PR.

### Version of `responses`

0.21.0

### Steps to Reproduce

```python
import json

from responses.matchers import _create_key_val_str, json_params_matcher


json_a = [{"a": "b"}]
json_b = [{"a": "b", "c": "d"}]
mock_request = Mock(body=json.dumps(json_b))
result = json_params_matcher(json_a)(mock_request)
print(result)

```

### Expected Result

`result` should be (False, "<reason>"), no error is thrown

### Actual Result

AttributeError is raised: 'list' object has no attribute 'keys'

beliaev-maksim:
closes #559 

I do not think we need to go for over-complicated comparisons of objects. User can extend functionality by applying custom matcher.
However, I do agree that lists should be supported as valid JSON

### DIFF
--- a/responses/matchers.py
+++ b/responses/matchers.py
@@ -97,16 +97,17 @@ def urlencoded_params_matcher(
 
 
 def json_params_matcher(
-    params: Optional[Dict[str, Any]], *, strict_match: bool = True
+    params: Optional[Union[Dict[str, Any], List[Any]]], *, strict_match: bool = True
 ) -> Callable[..., Any]:
     """Matches JSON encoded data of request body.
 
     Parameters
     ----------
-    params : dict
-        JSON data provided to 'json' arg of request or a part of it if used in
+    params : dict or list
+        JSON object provided to 'json' arg of request or a part of it if used in
         conjunction with ``strict_match=False``.
     strict_match : bool, default=True
+        Applied only when JSON object is a dictionary.
         If set to ``True``, validates that all keys of JSON object match.
         If set to ``False``, original request may contain additional keys.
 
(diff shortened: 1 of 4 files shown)


beliaev-maksim:
hi @chadwhitacre 

I think something is wrong with FOSSA pipeline. 22min for check is too much. I have already seen similar for other PRs. Can you please check?

responses/matchers.py
```diff
@@ -121,22 +122,30 @@ def json_params_matcher(
     def match(request: PreparedRequest) -> Tuple[bool, str]:
         reason = ""
         request_body = request.body
-        params_dict = params or {}
+        json_params = (params or {}) if not isinstance(params, list) else (params or [])
```
markstory, line 125:
```suggestion
        json_params = (params or {}) if not isinstance(params, list) else params
```

Do we need the second `or`? If params is a list that is falsey, it is already an empty list.

beliaev-maksim: you are right, thanks

chadwhitacre:
Glad this landed. Sorry for the FOSSA noise, we ended up [turning it off](https://github.com/getsentry/team-ospo/issues/4#issuecomment-1215091008) for now.

PR #574, opened 2022-08-23
Bugfix: Retry increment() method response object fix

lovmat:
Bugfix: Retry increment() method should be called with an urllib3.response.HTTPResponse response object

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -1028,7 +1028,7 @@ class RequestsMock(object):
                 retries = retries.increment(
                     method=response.request.method,  # type: ignore[misc]
                     url=response.url,  # type: ignore[misc]
-                    response=response,  # type: ignore[misc]
+                    response=response.raw,  # type: ignore[misc]
                 )
                 return self._on_request(adapter, request, retries=retries, **kwargs)
             except MaxRetryError:


beliaev-maksim:
Can you please add a test that will cover the difference and show what was fixed?

lovmat:
The type hints shows what it should be: https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/retry.py#L425

pyright fetches this:
```
$ pyright responses/__init__.py
Searching for source files
Found 1 source file
pyright 1.1.267

  responses/responses/__init__.py:1031:30 - error: Argument of type "Response | Unknown" cannot be assigned to parameter "response" of type "HTTPResponse | None" in function "increment"
    Type "Response | Unknown" cannot be assigned to type "HTTPResponse | None"
      Type "Response" cannot be assigned to type "HTTPResponse | None"
        "Response" is incompatible with "HTTPResponse"

```

There are also other type issues in the code so I'd recommend you start using `pyright`

markstory:
> The type hints shows what it should be: https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/retry.py#L425

Typehints can (and have been wrong in the past). How does the current implementation show up as a problem/bug for you?

lovmat:
Yes type hints can be wrong, but not in this case:  `Retry` is a class in `urllib3` - it wants an HTTPReponse object from `urllib3` - not from `requests`.

I'm overriding `increment` and using data from the response object. Noticed this when the tests failed but it worked when running the real thing (without `responses`)

beliaev-maksim:
@lovmat 
then can you please add these tests here to showcase and also prevent regression bugs in the future

PR #575, opened 2022-08-24
fix retry error

### LINKED ISSUE #572: Incorrect error raised when max retries are exceeded
### Describe the bug

When testing retry logic a `urllib3.exceptions.MaxRetryError` is raised instead of a `requests.exceptions.RetryError`.

### Additional context

_No response_

### Version of `responses`

0.21.0

### Steps to Reproduce

```python
from http.server import SimpleHTTPRequestHandler, ThreadingHTTPServer
from multiprocessing import Process
from tempfile import TemporaryDirectory
from time import sleep

import pytest
import requests
import responses
from requests.adapters import HTTPAdapter, Retry
from requests.exceptions import RetryError
from responses.registries import OrderedRegistry

_PORT = 8000

_SESSION = requests.session()
_SESSION.mount("http://", HTTPAdapter(max_retries=Retry(1, status_forcelist=[404])))


class CustomException(Exception):
    ...


def get(url: str):
    """System-under-test, catches retry errors and raises custom error"""
    try:
        return _SESSION.get(url)
    except RetryError:
        raise CustomException(url)


@responses.activate(registry=OrderedRegistry)
def test_get_responses():
    """Uses responses - fast, but has a bug"""
    url = f"http://localhost:{_PORT}/does_not_exist"
    rsp1 = responses.get(url, body="Error", status=404)
    rsp2 = responses.get(url, body="Error", status=404)

    with pytest.raises(CustomException):
        get(url)


def test_get_local_server(server: Process):
    """Uses a local server - slow, but correct"""
    url = f"http://localhost:{_PORT}/does_not_exist"
    with pytest.raises(CustomException):
        get(url)


### Fixtures to run a local server which serves 404s for everything but the root path


@pytest.fixture
def server():
    proc = Process(target=_run_server)
    proc.start()
    # Wait for the server to start, otherwise we may get a NewConnectionError as the connections
    # will be refused.
    sleep(1)
    yield proc
    proc.terminate()


def _run_server():
    """Creates a server that serves an empty directory 404'ing anything but the root path"""
    with TemporaryDirectory() as tempd:
        with ThreadingHTTPServer(
            ("", _PORT), lambda *args: SimpleHTTPRequestHandler(*args, directory=tempd)
        ) as httpd:
            httpd.serve_forever()
```

### Expected Result

Both tests pass.  More specifically that responses follows the logic ([original - v2.5.0](https://github.com/psf/requests/blob/v2.5.0/requests/adapters.py#L417-L424), [most recent version - v2.28.1](https://github.com/psf/requests/blob/v2.28.1/requests/adapters.py#L549-L565)) introduced in requests v2.5.0 ([the version that introduced the ability to use urllib3's Retry object with HTTPAdapters](https://github.com/psf/requests/blob/main/HISTORY.md#250-2014-12-01) ) that converts `urllib3.exceptions.MaxRetryError`s into `requests.exceptions` exceptions, specifically `ResponseError`s becoming `RetryError`s.

### Actual Result

The responses based test does not pass, as it raises `urllib3.exceptions.MaxRetryError` instead of a `requests.exceptions.RetryError`.

beliaev-maksim:
replace exception by RetryError according to requests implementation: https://github.com/psf/requests/blob/177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0/requests/adapters.py#L549

closes #572

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -27,6 +27,7 @@ from warnings import warn
 from requests.adapters import HTTPAdapter
 from requests.adapters import MaxRetryError
 from requests.exceptions import ConnectionError
+from requests.exceptions import RetryError
 
 from responses.matchers import json_params_matcher as _json_params_matcher
 from responses.matchers import query_string_matcher as _query_string_matcher
@@ -1031,9 +1032,17 @@ class RequestsMock(object):
                     response=response,  # type: ignore[misc]
                 )
                 return self._on_request(adapter, request, retries=retries, **kwargs)
-            except MaxRetryError:
+            except MaxRetryError as e:
                 if retries.raise_on_status:
-                    raise
+                    """Since we call 'retries.increment()' by ourselves, we always set "error"
+                    argument equal to None, thus, MaxRetryError exception will be raised with
+                    ResponseError as a 'reason'. That is why following 'if' condition is commented
+                    out. As reference see 'requests' implementation via:
+                    https://github.com/psf/requests/blob/
+                    177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0/requests/adapters.py#L549"""
+                    # if isinstance(e.reason, ResponseError):
+                    raise RetryError(e, request=request)
+
                 return response
         return response
 
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -1036,11 +1036,12 @@ class RequestsMock(object):
                 if retries.raise_on_status:
                     """Since we call 'retries.increment()' by ourselves, we always set "error"
                     argument equal to None, thus, MaxRetryError exception will be raised with
-                    ResponseError as a 'reason'. That is why following 'if' condition is commented
-                    out. As reference see 'requests' implementation via:
+                    ResponseError as a 'reason'. 
+                    
+                    Here we're emulating the `if isinstance(e.reason, ResponseError):`
+                    branch found at:
                     https://github.com/psf/requests/blob/
                     177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0/requests/adapters.py#L549"""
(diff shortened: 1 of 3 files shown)


1Mark:
LGTM

1Mark (approved):
LGTM

jared-duo (approved):
LGTM

responses/__init__.py
```diff
@@ -1031,9 +1032,17 @@ def _on_request(
...
                 if retries.raise_on_status:
-                    raise
+                    """Since we call 'retries.increment()' by ourselves, we always set "error"
+                    argument equal to None, thus, MaxRetryError exception will be raised with
+                    ResponseError as a 'reason'. That is why following 'if' condition is commented
+                    out. As reference see 'requests' implementation via:
+                    https://github.com/psf/requests/blob/
+                    177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0/requests/adapters.py#L549"""
+                    # if isinstance(e.reason, ResponseError):
+                    raise RetryError(e, request=request)
```
markstory, line 1044:
```suggestion
                    """Since we call 'retries.increment()' by ourselves, we always set "error"
                    argument equal to None, thus, MaxRetryError exception will be raised with
                    ResponseError as a 'reason'. 
                    
                    Here we're emulating the `if isinstance(e.reason, ResponseError):`
                    branch found at:
                    https://github.com/psf/requests/blob/
                    177dd90f18a8f4dc79a7d2049f0a3f4fcc5932a0/requests/adapters.py#L549"""
                    raise RetryError(e, request=request)
```

If we reword this we don't need to have the commented code and a comment on why we have commented out code.

beliaev-maksim: @markstory thank you!

PR #581, opened 2022-09-01
[BETA] replay responses from toml files.

### LINKED ISSUE #40: Add expected response from a file
In your [introductory blog post](http://cramer.io/2014/05/20/mocking-requests-with-responses/) you have the last example with a lot of such calls:

```
responses.add(
            responses.GET, 'http://jenkins.example.com/job/server/2/logText/progressiveHtml/?start=0',
            match_querystring=True,
            adding_headers={'X-Text-Size': '7'},
            body='Foo bar')
```

This is clearly a lack of the feature and the workaround because it is barely practical to stuff one's test suite with big response strings. And a full response file can contain all of status, headers and response body. It can look like this (py3):

```
from io import BytesIO
from http.client import HTTPResponse

import responses


class FakeSocket():

  _stream = None


  def __init__(self, s):
    self._stream = BytesIO(s)

  def makefile(self, *args, **kwargs):
    return self._stream


def parse_response(request_string):
  response = HTTPResponse(FakeSocket(request_string))
  response.begin()
  return response

class ResponseMock(responses.RequestsMock):

  def add_from_file(self, method, url, filename, match_querystring = False, stream = False):  
    with open(filename, 'rb') as f:
      r = parse_response(f.read())

    content_type = r.getheader('content-type', 'text/plain')

    self.add(method, url, r.read(), match_querystring, r.status, r.getheaders(), stream, content_type)
```

Response file can look like this:

```
HTTP/1.1 200 OK
Date: Sat, 22 Nov 2014 12:00:40 GMT
Content-Type: text/xml; charset="utf-8"
Connection: close

Some <em>HTML</em> goes here.
```

beliaev-maksim:
closes #40

### DIFF
--- a/responses/tests/test_recorder.py
+++ b/responses/tests/test_recorder.py
@@ -102,11 +102,11 @@ class TestRecord:
 
 class TestReplay:
     def teardown(self):
-        self.out_file = Path("out.toml")
-        if self.out_file.exists():
-            self.out_file.unlink()
+        out_file = Path("out.toml")
+        if out_file.exists():
+            out_file.unlink()
 
-        assert not self.out_file.exists()
+        assert not out_file.exists()
 
     def test_add_from_file(self):
         with open("out.toml", "w") as file:
--- a/responses/tests/test_recorder.py
+++ b/responses/tests/test_recorder.py
@@ -114,22 +114,29 @@ class TestReplay:
 
         @responses.activate
         def run():
+            responses.patch("http://httpbin.org")
             responses._add_from_file(file_path="out.toml")
-            assert responses.registered()[0].url == "http://example.com:8080/404"
+            responses.post("http://httpbin.org/form")
+
+            assert responses.registered()[0].url == "http://httpbin.org/"
+            assert responses.registered()[1].url == "http://example.com:8080/404"
             assert (
-                responses.registered()[1].url == "http://example.com:8080/status/wrong"
+                responses.registered()[2].url == "http://example.com:8080/status/wrong"
             )
-            assert responses.registered()[2].url == "http://example.com:8080/500"
-            assert responses.registered()[3].url == "http://example.com:8080/202"
+            assert responses.registered()[3].url == "http://example.com:8080/500"
+            assert responses.registered()[4].url == "http://example.com:8080/202"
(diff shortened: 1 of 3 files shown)


beliaev-maksim:
@markstory @chadwhitacre 
historical moment :)
we went down from more than 70 issues last year, to 0, once this PR will be merged and new package is released. Implemented lots of new feature and bug fixes for them :D 
Making the library №1 for mocking out requests

chadwhitacre:
Awesome progress, @beliaev-maksim! 😍

responses/tests/test_recorder.py
```diff
@@ -48,47 +97,39 @@ def run():
...
+        assert data == get_data(httpserver.host, httpserver.port)
+
+
+class TestReplay:
+    def teardown(self):
+        self.out_file = Path("out.toml")
+        if self.out_file.exists():
+            self.out_file.unlink()
+
+        assert not self.out_file.exists()
```
markstory, line 109:
```suggestion
        out_file = Path("out.toml")
        if out_file.exists():
            out_file.unlink()

        assert not out_file.exists()
```

Do we need to retain the attribute here?

responses/tests/test_recorder.py
```diff
@@ -48,47 +97,39 @@ def run():
...
+
+    def test_add_from_file(self):
```
markstory, line 111: Do you think its useful to have test coverage for how fixture files interact with responses that were defined before/after the file was loaded?

responses/tests/test_recorder.py
beliaev-maksim: true

responses/tests/test_recorder.py
beliaev-maksim: generally, since we only append `Responses`, it should be fine as is.
But if I add couple of lines in the test it for sure will not harm :)

markstory:
Great job on this @beliaev-maksim :clap: I think we're ready for a release now.

beliaev-maksim:
@markstory 
yes,
all the rest PRs we can push to 0.23

sanderwilrycx:
Any thoughts on when the release will be? We are eagerly awaiting to test the recording functionality!

PR #583, opened 2022-09-01
apply partialmethod to shortcuts

beliaev-maksim:
I think this is more pythonic way of doing shortcuts

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -12,7 +12,7 @@
 * Add support for the `loose` version of `json_params_matcher` via named argument `strict_match`. See #551
 * Add lists support as JSON objects in `json_params_matcher`. See #559
 * Added project links to pypi listing.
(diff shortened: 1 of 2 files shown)


CHANGES
```diff
@@ -9,6 +9,7 @@
 * Add support for the `loose` version of `json_params_matcher` via named argument `strict_match`. See #551
 * Add lists support as JSON objects in `json_params_matcher`. See #559
 * Added project links to pypi listing.
+* `delete`, `get`, `head`, `options`, `patch`, `post`, `put` shortcuts are redone using `functools.partialmethod`.
```
markstory, line 12:
```suggestion
* `delete`, `get`, `head`, `options`, `patch`, `post`, `put` shortcuts are now implemented using `functools.partialmethod`.
```

PR #587, opened 2022-09-02
add docstrings to OrderedRegistry

beliaev-maksim:


responses/registries.py
```diff
@@ -75,9 +75,32 @@ def replace(self, response: "BaseResponse") -> "BaseResponse":
 
 
 class OrderedRegistry(FirstMatchRegistry):
+    """Registry where `Response` objects are dependent on the insertion order and invocation index.
+
+    Current registry applies the rule of first in - first out. Responses should be invoked in
```
markstory, line 80:
```suggestion
    OrderedRegistry applies the rule of first in - first out. Responses should be invoked in
```

responses/registries.py
```diff
@@ -75,9 +75,32 @@ def replace(self, response: "BaseResponse") -> "BaseResponse":
...
+    the same exact order in which they were added to the registry. Otherwise, error is returned.
```
markstory, line 81:
```suggestion
    the same order in which they were added to the registry. Otherwise, an error is returned.
```

responses/registries.py
```diff
@@ -75,9 +75,32 @@ def replace(self, response: "BaseResponse") -> "BaseResponse":
...
+        Parameters
+        ----------
+        request : PreparedRequest
+            Request that was caught by the custom adapter.
+
+        Returns
+        -------
+        Tuple[Optional["BaseResponse"], List[str]]
+            Matched `Response` object and empty list in case of match.
+            Otherwise, None and a list with reasons why it doesn't match.
```
markstory, line 101:
```suggestion
        """Find the next registered `Response` and check if it matches the request.

        Search is performed by taking the first element of the registered responses list
        and removing this object (popping from the list).

        Parameters
        ----------
        request : PreparedRequest
            Request that was caught by the custom adapter.

        Returns
        -------
        Tuple[Optional["BaseResponse"], List[str]]
            Matched `Response` object and empty list in case of match.
            Otherwise, None and a list with reasons for not finding a match.
```

beliaev-maksim:
@markstory
Thank you!

Issue #556, opened 2022-05-27
responses 0.21.0 broke mypy check


ondratu:
### Describe the bug

We use response in our tests, and from responses 0.21.0 we've got this error:
```
$mypy tests/test_download.py 
venv/lib/python3.10/site-packages/mypy/typeshed/stdlib/unittest/mock.pyi: error: Source file found twice under different module names: "unittest.mock" and "mypy.typeshed.stdlib.unittest.mock"
Found 1 error in 1 file (errors prevented further checking)
```

Returning  mypy version did not resolve problem, but returning responses to 0.20.0 does.

I try to find problem in your changes, and it could be probably this line in __init__.py:

```
from mypy.typeshed.stdlib.unittest.mock import _patcher as _mock_patcher
```

### Version of `responses`

0.21.0

### Steps to Reproduce

```python
# content of crash.py

import responses
import unittest.mock
```

and than run
```
mypy crash.py
```

### Expected Result

Success: no issues found in 1 source file

### Actual Result

venv/lib/python3.10/site-packages/mypy/typeshed/stdlib/unittest/mock.pyi: error: Source file found twice under different module names: "unittest.mock" and "mypy.typeshed.stdlib.unittest.mock"
Found 1 error in 1 file (errors prevented further checking)

markstory:
I'm not able to reproduce the problem you're having

```
ᐉ pyenv install 3.10.1
python-build: use openssl@1.1 from homebrew
python-build: use readline from homebrew
Downloading Python-3.10.1.tar.xz...
-> https://www.python.org/ftp/python/3.10.1/Python-3.10.1.tar.xz
Installing Python-3.10.1...
python-build: use readline from homebrew
python-build: use zlib from xcode sdk
Installed Python-3.10.1 to /Users/markstory/.pyenv/versions/3.10.1

ᐉ pyenv local 3.10.1
ᐉ pip install responses
Collecting responses
  Downloading responses-0.21.0-py3-none-any.whl (45 kB)
     |████████████████████████████████| 45 kB 4.0 MB/s
Collecting requests<3.0,>=2.0
  Downloading requests-2.27.1-py2.py3-none-any.whl (63 kB)
     |████████████████████████████████| 63 kB 8.5 MB/s
Collecting urllib3>=1.25.10
  Downloading urllib3-1.26.9-py2.py3-none-any.whl (138 kB)
     |████████████████████████████████| 138 kB 16.2 MB/s
Collecting idna<4,>=2.5
[...]

PR #593, opened 2022-10-12
Fix annotations of CallList

ziima:
With 0.22.0, mypy started to complain about constructions `rsps.calls[0].request` with an error
```
Item "List[Call]" of "Union[Call, List[Call]]" has no attribute "request"  [union-attr]
```
because it wasn't able to differentiate between the return types. I added `overload` annotation to handle that.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,8 @@
+0.23.0
+------
+
+* Fix type annotations of `CallList`.
+
 0.22.0
 ------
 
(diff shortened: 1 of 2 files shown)


beliaev-maksim:
hi, thanks

please add new version in changelog and add your item there

ziima:
Done.

beliaev-maksim (approved):

markstory:
Thank you 🎉

andersk:
When you get a chance, it’d be great to get a new PyPI release, since none of the existing releases work with mypy (#552 for ≤ 0.20.0, #556 for 0.21.0, this for 0.22.0) and current Git works.

beliaev-maksim:
closes #610

Issue #595, opened 2022-10-12
Redundant runtime dependency on types-toml


stanislavlevin:
### Describe the bug

With f431d49021252657e55253c18f0f8493c61e6ae6 `types-toml` is mandatory runtime dependency, while
it's actually typing stub.

### Version of `responses`

0.22.0

### Steps to Reproduce

```python
# your code goes here

```

### Expected Result

There should be no dependency on typing stuff at runtime.

### Actual Result

 There is runtime dependency on types-toml.

beliaev-maksim:
I am -1 on removing it from run time. Users that have installed responses should not bother themselves to install types for toml to work with our project and do a type checking

@markstory 
thoughts?

stanislavlevin:
This can be done with extras (place all the required typing stuff there). If somebody wants to do type checking they can install it (e.g. `responses[typing]`) explicitly. But if a requirement is strict it would be pulled unconditionally ( i.e. I have no choice).

markstory:
Why is this a problem? Is it because you would like to avoid downloading the additional package?

beliaev-maksim:
moved to tomli

PR #596, opened 2022-10-12
Use tomllib/tomli/tomli-w instead of unmaintained toml [WIP]

mgorny:
Replace the `toml` dependency that is unmaintained (last release in 2020) and does not implement TOML 1.0 with more modern libraries. Use the built-in `tomllib` module to load .toml files on Python 3.11, and its drop-in replacement `tomli` package on older versions of Python. For writing .toml files, use the `tomli-w` package.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -25,7 +25,7 @@ from typing import Union
 from typing import overload
 from warnings import warn
 
-import toml as _toml
+import tomli as _toml
 from requests.adapters import HTTPAdapter
 from requests.adapters import MaxRetryError
 from requests.exceptions import ConnectionError
@@ -770,7 +770,7 @@ class RequestsMock(object):
     put = partialmethod(add, PUT)
 
     def _add_from_file(self, file_path: "Union[str, bytes, os.PathLike[Any]]") -> None:
-        with open(file_path) as file:
+        with open(file_path, "rb") as file:
             data = _toml.load(file)
 
         for rsp in data["responses"]:
--- a/responses/tests/test_recorder.py
+++ b/responses/tests/test_recorder.py
@@ -1,7 +1,8 @@
 from pathlib import Path
 
 import requests
-import toml
+import tomli as _toml
+import tomli_w
 
 import responses
 from responses import _recorder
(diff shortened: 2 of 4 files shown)


mgorny:
There's one problem with this: right now it fails because it attempts to serialize `None` that has no valid representation in TOML. Apparently the buggy `toml` module either ignored keys with `None` as value or serialized it as string `"None"` (sic!).

responses/tests/test_recorder.py
```diff
@@ -1,7 +1,9 @@
+import sys
```
beliaev-maksim, line 1:
```suggestion
```

responses/__init__.py
```diff
@@ -1,6 +1,7 @@
 import inspect
 import json as json_module
 import logging
+import sys
```
beliaev-maksim, line 4:
```suggestion
```

beliaev-maksim (changes requested):

beliaev-maksim:
makes sense to switch. Thanks for the PR

at the time of writing the feature I chose `toml` because of the wider usage by the community. However, if it is unmaintained there is no reason not to go for `tomli`

will also resolve #595

mgorny:
> at the time of writing the feature I chose `toml` because of the wider usage by the community.

I'm surprised at that but I suppose there's still work to be done there. I think most of the major projects (black, flit, mypy, pytest, setuptools, tox to list just a few) have switched to `tomli` already. At least a quick glance it seems that `tomli` has more revdeps in Gentoo than `toml`.

mgorny:
See https://github.com/getsentry/responses/pull/596#issuecomment-1276008203. The problem is that the input is broken right now, and I'm looking for guidance what fix would you prefer for it.

beliaev-maksim:
> Apparently the buggy `toml` module either ignored keys with `None`

it is not buggy, but done on purpose
I just went over discussions on https://github.com/toml-lang/toml/issues/921
also, read threads in `tomli-w` and `toml`

according to toml-lang the value with null should not be present in the file, thus, it should be filtered out in languages where it is present eg python. So, we need to filter all nulls out.

```python
def remove_nones(d):
    if isinstance(d, dict):
        return {k: remove_nones(v) for k, v in d.items() if v is not None}
    return d
```

and then dump the result

mgorny:
Thanks for the snippet. Done that now and the tests seem to pass.

markstory:
Thank you 🎉

mgorny:
Thanks!

mgorny:
> we will add support for 3.11 once it is released please remove it so far

Python 3.11.0 has been released on Oct 24th.

PR #606, opened 2022-11-15
coverage

beliaev-maksim:
replace function's body placeholders with docstrings to increase coverage


beliaev-maksim:
@markstory 
propose to merge PRs only with 100% coverage, since #593 dropped it by 0.07%

PR #613, opened 2022-12-22
rerun pre-commit on blacken docs

beliaev-maksim:


beliaev-maksim:
@markstory can you please check and merge, otherwise CI fails on master

PR #618, opened 2023-01-26
Setup CodeQL

mdtro:
Basic setup for CodeQL to scan on PRs targeting master and once a week.

### DIFF
--- a/.github/codeql/codeql-config.yml
+++ b/.github/codeql/codeql-config.yml
@@ -1,0 +1,4 @@
+name: "Sentry CodeQL Config"
+
+paths-ignore:
+  - '**/tests/**'
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,54 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  schedule:
+    - cron: '36 14 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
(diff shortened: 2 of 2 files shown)


beliaev-maksim:
What is the value of it?
What could be identified and how is it reported?

mdtro:
> What is the value of it? What could be identified and how is it reported?

@beliaev-maksim good questions! Apologies I did not include them in my original PR description.

**What is the value of it?** 

CodeQL is used to perform security analysis on the codebase. It will use known insecure patterns to identify similar issues in `responses`. 

Generally, CodeQL has a low false positive rate. Implementing this tool helps us be good stewards of a relatively popular library package to ensure we are not introducing security vulnerabilities into codebases that depend on us. 🙂 

**What could be identified and how is it reported?**
I've extracted the CodeQL detections relevant to Python here: https://gist.github.com/mdtro/2017a2f2b19748a543442c9b35ab78ac

The configuration I'm proposing here uses queries just in the `code-scanning` suite as a conservative measure, but we can always expand to the `security-extended` or others in the future.

Detections will be reported inline with the potentially vulnerable code on PRs and can be resolved/dismissed directly in the PR. They'll also show up under the _Security_ tab and are visible to those with the necessary permissions. Detections in PRs are visible to anyone.

markstory:
Seems fine to me. For context @beliaev-maksim CodeQL is being applied across many of the actively maintained repositories at sentry.

mdtro:
I do like keeping this in the PR as it closes the feedback loop for contributors. 🙂 

I will adjust this PR to exclude the `tests` directory to lower the noise ratio.

beliaev-maksim:
@mdtro can you please delete branch?

PR #631, opened 2023-03-10
remove tomli import

### LINKED ISSUE #630: responses now seems to require tomllib and tomli but it does not get installed via pip.
### Describe the bug

I have a requirements.txt file that uses moto (ver. 3.0.6) and this has responses as one of it's dependencies. With the latest release, version 0.23.0 is collected as part of pip installing the requirements file. But when running unit tests which involve moto, the responses pkg is inevitably used and the tests are **failing** because we get a ModuleNotFound error for "tomli" and tomllib. 

Example images of errors are shown below:

### Additional context

Screen Shot 2023-03-10 at 12 35 23 PM

### Version of `responses`

0.23.0

### Steps to Reproduce

```python
# your code goes here
python3 -m venv temp_venv/   # Create a fresh venv.
. temp_venv/bin/activate   # Activate. 
pip install moto==3.0.6    # will install responses==0.23.0
# Run any unit test involving moto library --> should fail due to missing tomli / tomllib.
```

### Expected Result

Tests involving moto should not fail because dependencies for the pkg should be installed properly. 

### Actual Result

Tests failing because of these missing dependencies.

beliaev-maksim:
closes #630


Jylanthas:
thanks. this worked for me

PR #641, opened 2023-05-28
Record matcher

### LINKED ISSUE #639: _recorder - matching http headers as well as URL
For my project, I need to mock responses from an API where some of the parameters are included in a HTTP header, so for example I would make a request to `http://api.com/projects` with a http header set to `user-id: joebloggs`

Currently I am recording the responses by doing:

```python
@_recorder.record(filepath="out.yaml")
def test_something():
    ...
```

and then I'm loading the yaml file like this:

```python
@responses.activate
def test_something():
    responses.add_from_file("out.yaml")
    ...
```

Is it possible to set up the registry to match on this HTTP header, and also save this header in the .yaml files that it generates?

beliaev-maksim:
closes #639 

Current PR proposes the implementation of recorder for matchers into yaml (or any other compatible format)

### DIFF
--- a/responses/tests/test_recorder.py
+++ b/responses/tests/test_recorder.py
@@ -242,7 +242,7 @@ class TestRecord:
 
 class TestReplay:
     def setup(self):
-        self.out_file = Path("response_record.yaml")
+        self.out_file = Path("response_record")
 
     def teardown(self):
         if self.out_file.exists():
--- a/responses/tests/test_recorder.py
+++ b/responses/tests/test_recorder.py
@@ -21,42 +21,141 @@ def get_data(host, port):
         "responses": [
             {
                 "response": {
-                    "method": "GET",
-                    "url": f"http://{host}:{port}/404",
+                    "auto_calculate_content_length": False,
                     "body": "404 Not Found",
+                    "content_type": "text/plain",
+                    "matchers": [
+                        {
+                            "query_param_matcher": {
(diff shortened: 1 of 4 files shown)


responses/tests/test_recorder.py
```diff
@@ -21,42 +21,141 @@ def get_data(host, port):
...
+                                "args": {
+                                    "headers": {
+                                        "Accept": "*/*",
+                                        "Accept-Encoding": "gzip, deflate",
+                                        "Connection": "keep-alive",
+                                        "User-Agent": "python-requests/2.30.0",
+                                    },
+                                    "strict_match": False,
+                                },
+                                "matcher_import_path": "responses.matchers",
```
beliaev-maksim, line 45: hi @markstory 

pls have a look on the draft of the recorder.
I have a question/concern how to proceed with replay

considering that we need to support custom matchers, we probably want to import matchers as well.
I found a way how to extract the location of the mather. however, then we need to do import during runtime and local to each function, which does not sound really smart

what could be the options?

I think asking users to import required matchers defeats the purpose of record/replay

markstory: Hrm, importing userland code is possible but it could get messy. I think we're also taking a risk that userland matcher parameters will be 'serializable' into JSON, which isn't guaranteed. We could require that in order to use recording :shrug:

PR #648, opened 2023-06-24
move unbound_send outside of the function

beliaev-maksim:
Moving `unbound_on_send` outside of the function will allow users to override the method if they will need to without a need to rewrite the entire `start` method

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -38,10 +38,8 @@ from responses.registries import FirstMatchRegistry
 
 try:
     from typing_extensions import Literal
-    from typing_extensions import Protocol
 except ImportError:  # pragma: no cover
     from typing import Literal  # type: ignore  # pragma: no cover
-    from typing import Protocol
 
 try:
     from requests.packages.urllib3.response import HTTPResponse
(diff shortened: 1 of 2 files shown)


responses/__init__.py
```diff
@@ -1091,14 +1094,18 @@ def _on_request(
...
-            # another decorated function
-            return
-
-        def unbound_on_send(
+    def unbound_on_send(
+        self,
+    ) -> """Callable[  # noqa: F821
+        [
+            Arg(HTTPAdapter, "adapter"),
+            Arg(PreparedRequest, "request"),
```
beliaev-maksim, line 1102: @asottile-sentry can you please may be suggest here?

For some reason flake8 complains that
```
responses/__init__.py:1101:30: F821 undefined name 'adapter'
responses/__init__.py:1102:34: F821 undefined name 'request'
```

but `Arg` from mypy extensions takes just a name of the positional argument. So, what is not located by `flake8` ?

asottile-sentry: use a `Protocol` -- `Arg` / etc. are deprecated mypy-specific types and will not work with other type checkers

beliaev-maksim: @asottile-sentry gotcha, thanks

fixed in the next commit

beliaev-maksim:
@markstory while working on the bug for `moto` I spotted that we always use the same _real_send. However, there might be a use case where we will use `unbound_on_send` in a nested responses mock as a real adapter.
then user can intercept real send from a mock (eg in moto) and override it with responses.unbound_on_send 

a bit hard to explain. 
here is more info and examples: https://github.com/getmoto/moto/issues/6417

PR #655, opened 2023-07-25
please fix this error

JohnVillalovos:
So delete with content doesn't work 🥲


JohnVillalovos:
So after further investigation this seems a `requests` issue where if it is a 204 (No Content) then it sets the expected length to 0.

PR #658, opened 2023-07-27
chore: UserWarning if body provided for unsupported status

JohnVillalovos:
After the change to urllib3 v2, certain status codes are not expected to have any content returned in the response. If they do have content then requests (via urllib3) will cause an error like:
  `requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(18 bytes read, -18 more expected)', IncompleteRead(18 bytes read, -18 more expected))`

This can be confusing to users using responses who had set a body in previous versions of urllib3 without any issues. Emit a `UserWarning` to help users remedy the issue.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -567,6 +567,17 @@ class Response(BaseResponse):
                 "stream argument is deprecated. Use stream parameter in request directly",
                 DeprecationWarning,
             )
+        # Check for responses that shouldn't include a body
+        if body:
+            if status in (204, 304) or 100 <= status < 200:
+                warn_msg = (
+                    f"Providing a body/json value with status == {status} may cause a "
+                    f"`requests.exceptions.ChunkedEncodingError: ('Connection broken: "
+                    f"IncompleteRead({len(body)} bytes read, -{len(body)} more "
+                    f"expected)', IncompleteRead({len(body)} bytes read, -{len(body)} "
+                    f"more expected))` error."
+                )
+                warn(warn_msg, category=UserWarning, stacklevel=3)
 
         self.stream: Optional[bool] = stream
         self.content_type: str = content_type  # type: ignore[assignment]
--- a/responses/tests/test_responses.py
+++ b/responses/tests/test_responses.py
@@ -2574,3 +2574,39 @@ def test_request_object_attached_to_exception():
 
     run()
     assert_reset()
+
+
+def test_status_204_unexpected_body_warning():
+    @responses.activate
+    def run():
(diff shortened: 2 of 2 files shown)


beliaev-maksim:
I am -1 on this

This is default behaviour of urllib and reqiests which has nothing to do with this lib.

Users must read upgrade manuals for urllib and requsts before bumping those

beliaev-maksim:
@JohnVillalovos if the error message is unclear, I would recommend to open an issue against urllib

since if this error will hit in production, responses will not help with warning

JohnVillalovos:
> @JohnVillalovos if the error message is unclear, I would recommend to open an issue against urllib
> 
> since if this error will hit in production, responses will not help with warning

It was an error that we were only seeing in testing. As we set the response body while also setting status 204

There were no issues in real code running.

My PR to resolve this issue in our project: https://github.com/python-gitlab/python-gitlab/pull/2627

JohnVillalovos:
To clarify. It was only something we were seeing when doing unit testing using `responses`. As instructing `responses` to have a status code of 204 and also providing a body will then cause `urllib3`/`requests` to generate an exception. So not something seen in the real world but only seen in unit testing with `responses`.

beliaev-maksim:
because in reality backend should not respond with 204 and content in the response. If it does, then it is an issue in the backend.

responses emulate real request. If in the real request you get 204 with content attached. Then please file an issue with example with and without responses and we will see how to handle it

JohnVillalovos:
I feel like we are not understanding each other.

So my purpose for this PR was to help the user of `responses` when they use it incorrectly:
  * User incorrectly sets up or has previously set up `responses` (what happened to our project) to provide a return body/json value for a status code like 204
  * Now user when running their unit tests are getting `ChunkedEncodingError` and are confused.
  * This PR provides a warning to the user helping to explain what is wrong so hopefully they can fix it more easily.

The main beneficiaries of this are going to be those people who were blissfully unaware they had done something wrong by using a body/json value with a status code like 204. Then they attempt to upgrade and get a bunch of `ChunkedEncodingError` and don't know why.

This is only about a user doing something wrong using `responses` and trying to help them.

beliaev-maksim:
I still think it is not something wrong with responses. It is general concept of trying to mock wrong data.

for this purpose we are working on responses recorder that will record real responses to files. Then users even do not need to think about it.
see usage example: https://github.com/canonical/gh-jira-sync-bot/blob/main/tests/unit/test_server.py

Issue #657, opened 2023-07-26
Overly strict version constraint on urllib3


damian-rp:
Hello team,

I see that in yesterday's release, you have changed `responses` to require v2.0 of urllib3 or higher.

This is causing us some dependency conflicts, as our projects use other packages that are still running on the v1.0 branch.

I see that urllib3 v1.0 is still being actively maintaned. Is there a specific reason that `responses` can no longer use it? I had a look at the discussions in #635 and #636, which discuss compatibility issues with v3.0, but I don't understand why the minimum version was changed here.

Since `responses` is of course a testing library that we don't install in production, I would opine that it should keep version constraints as loose as possible, so that it accurately tests the version(s) of other packages that would be installed in a production environment. 🙂 Is it possible to get this changed?

JohnVillalovos:
I have proposed PR #659 to allow older versions of `urllib3`. Not sure if it will be accepted or not.

bearrito:
Same. This library is used by Moto which I think is the standard test library for AWS services. That library relies on boto. Boto has a hard constraint for urlib < 1.27

PR #659, opened 2023-07-28
chore: allow urllib3>=1.25.10

JohnVillalovos:
Allow use of urllib3>=1.25.10. By default it will use urllib3>=2 as that is the latest version. But this will allow people to pin urllib3<2 and still use the latest version of requests and responses.

Closes: #657


setup.py
```diff
@@ -18,7 +18,7 @@
 
 install_requires = [
     "requests>=2.30.0,<3.0",
-    "urllib3>=2.0.0,<3.0",
+    "urllib3>=1.25.10",
```
markstory, line 21:
```suggestion
    "urllib3>=1.25.10,<3.0",
```

We should keep the upper bound. Future versions of urllib3 may require changes in responses.

JohnVillalovos: Thanks. Done.

beliaev-maksim:
please add a matrix of tests in CI. We need to test on both 1.x and 2.x

otherwise we give a promise that we cannot guarantee

JohnVillalovos:
> please add a matrix of tests in CI. We need to test on both 1.x and 2.x
> 
> otherwise we give a promise that we cannot guarantee

Thanks.

Done, I think. The CI doesn't run for me yet so I'm not 100% certain.

JohnVillalovos:
@beliaev-maksim or @markstory   Can one of you enable the workflow to run? Thanks.

JohnVillalovos:
> Done, I think. The CI doesn't run for me yet so I'm not 100% certain.

I was able to run the CI over in my fork. And I got the tests to pass. https://github.com/JohnVillalovos/responses/pull/2

Seluj78 (approved):

Issue #651, opened 2023-07-13
Feature request: Allow regex-matching for header values


spreeni:
I would like to match the expected format of a request-signature, without necessarily knowing the exact signature hash. To enable this or similar use cases, it would be nice to be able to supply compiled regex for the `matchers.header_matcher` similar as it is possible for the url.

In general, this could be a nice feature for all other matchers as well.

**Example:**
If I want to for example test that an outgoing request to an external API is supplied with a valid signature and a Bearer token. Currently, I'd implement a custom matcher for this. But it would be nice to test header fields optionally with regex directly in the `header_matcher`.
```python
responses.get(
    url,
    status=200,
    json=mock_data,
    match=[
        matchers.header_matcher(
            {
                "Message-Signature": re.compile(r'signature="\S+",created=\d+'),
                "Authorization": "Bearer API_TOKEN",
            },
            strict_match=False
        ),
        signature_matcher,
    ],
)

```
Ps: I love `responses`, thanks for the great work :)

beliaev-maksim:
please provide an example of the expected syntax and use case

spreeni:
I added an example, please see if it helps clarify what I meant.

spreeni:
@beliaev-maksim Hey sorry, I missed the notification somehow. Yes, I found custom matchers and that is how I implemented it myself now. But maybe as a convenience feature it would be nice to have this functionality directly within the matchers. It could make a fairly common approach more readable.

beliaev-maksim:
@spreeni can you fire a PR with implementation proposal based on your custom matcher?

geetptl:
@beliaev-maksim can I take this?

beliaev-maksim:
@geetptl your contribution will be warmly welcomed!

Please check Contribution section https://github.com/getsentry/responses#contributing

geetptl:
Think I'll need to add one more test like `test_request_matches_headers_regex_strict`. Might need a little more work, I'll get back soon.

PR #663, opened 2023-08-09
[#651] regex compatability for headers matching

geetptl:
Attempt at #651.

- Changes in `header_matcher` function.
- Instead of using simple list equality, it now checks for either regex or simple string check.
- Added tests `test_request_matches_headers_regex`, `test_request_matches_headers_regex_strict_match`

### DIFF
--- a/responses/matchers.py
+++ b/responses/matchers.py
@@ -403,18 +403,22 @@ def header_matcher(
     :return: (func) matcher
     """
 
-    def __compare_with_regex(request_headers: Union[Dict[Any, Any], Any]) -> bool:
-        checked_request_headers = {k: 1 for k in request_headers.keys()}
+    def _compare_with_regex(request_headers: Union[Dict[Any, Any], Any]) -> bool:
+        if strict_match and len(request_headers) != len(headers):
+            return False
+
         for k, v in headers.items():
             if request_headers.get(k) is not None:
-                checked_request_headers[k] = 0
                 if isinstance(v, re.Pattern):
                     if re.match(v, request_headers[k]) is None:
                         return False
                 else:
                     if not v == request_headers[k]:
                         return False
-        return True if not strict_match else sum(checked_request_headers.values()) == 0
+            elif strict_match:
+                return False
+
+        return True
 
     def match(request: PreparedRequest) -> Tuple[bool, str]:
         request_headers: Union[Dict[Any, Any], Any] = request.headers or {}
@@ -423,7 +427,7 @@ def header_matcher(
             # filter down to just the headers specified in the matcher
             request_headers = {k: v for k, v in request_headers.items() if k in headers}
 
-        valid = __compare_with_regex(request_headers)
+        valid = _compare_with_regex(request_headers)
 
         if not valid:
             return False, "Headers do not match: {} doesn't match {}".format(
--- a/responses/matchers.py
+++ b/responses/matchers.py
@@ -1,4 +1,5 @@
 import json as json_module
+import re
 from json.decoder import JSONDecodeError
 from typing import Any
 from typing import Callable
@@ -402,6 +403,19 @@ def header_matcher(
     :return: (func) matcher
     """
 
+    def __compare_with_regex(request_headers: Union[Dict[Any, Any], Any]) -> bool:
+        checked_request_headers = {k: 1 for k in request_headers.keys()}
+        for k, v in headers.items():
+            if request_headers.get(k) is not None:
+                checked_request_headers[k] = 0
+                if isinstance(v, re.Pattern):
+                    if re.match(v, request_headers[k]) is None:
+                        return False
+                else:
+                    if not v == request_headers[k]:
+                        return False
+        return True if not strict_match else sum(checked_request_headers.values()) == 0
+
     def match(request: PreparedRequest) -> Tuple[bool, str]:
         request_headers: Union[Dict[Any, Any], Any] = request.headers or {}
 
@@ -409,7 +423,7 @@ def header_matcher(
             # filter down to just the headers specified in the matcher
             request_headers = {k: v for k, v in request_headers.items() if k in headers}
 
-        valid = sorted(headers.items()) == sorted(request_headers.items())
+        valid = __compare_with_regex(request_headers)
 
         if not valid:
             return False, "Headers do not match: {} doesn't match {}".format(
--- a/responses/tests/test_matchers.py
+++ b/responses/tests/test_matchers.py
@@ -818,7 +818,51 @@ def test_request_matches_headers_regex():
     assert_reset()
 
 
-def test_request_matches_headers_regex_strict_match():
+def test_request_matches_headers_regex_strict_match_regex_failed():
+    @responses.activate
+    def run():
+        url = "http://example.com/"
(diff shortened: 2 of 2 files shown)


responses/matchers.py
```diff
@@ -402,14 +403,27 @@ def header_matcher(
...
+
     def match(request: PreparedRequest) -> Tuple[bool, str]:
         request_headers: Union[Dict[Any, Any], Any] = request.headers or {}
 
         if not strict_match:
             # filter down to just the headers specified in the matcher
             request_headers = {k: v for k, v in request_headers.items() if k in headers}
 
-        valid = sorted(headers.items()) == sorted(request_headers.items())
+        valid = __compare_with_regex(request_headers)
```
beliaev-maksim, line 426:
```suggestion
        valid = _compare_with_regex(request_headers)
```

responses/matchers.py
```diff
@@ -402,14 +403,27 @@ def header_matcher(
     :return: (func) matcher
     """
 
+    def __compare_with_regex(request_headers: Union[Dict[Any, Any], Any]) -> bool:
+        checked_request_headers = {k: 1 for k in request_headers.keys()}
+        for k, v in headers.items():
+            if request_headers.get(k) is not None:
```
beliaev-maksim, line 409: I assume this will crash if I add a header into my matcher that is not present in the request, or ?

beliaev-maksim (changes requested):

responses/matchers.py
```diff
@@ -402,14 +403,27 @@ def header_matcher(
...
+                checked_request_headers[k] = 0
+                if isinstance(v, re.Pattern):
+                    if re.match(v, request_headers[k]) is None:
+                        return False
+                else:
+                    if not v == request_headers[k]:
+                        return False
+        return True if not strict_match else sum(checked_request_headers.values()) == 0
```
beliaev-maksim, line 417: can it be refactored as?
```suggestion
        for k, v in headers.items():
            if request_headers.get(k) is not None:
                if isinstance(v, re.Pattern):
                    if re.match(v, request_headers[k]) is None:
                        return False
                else:
                    if not v == request_headers[k]:
                        return False
            elif strict_match:
                return False

```

responses/matchers.py
geetptl: I'll add this in the following commit.

responses/matchers.py
geetptl: This will work only when there's a field in `header` which doesn't match anything in `request_header`. If we have a field in `request_header` that is not present in `header`, this suggestion won't capture that.

Although this is a good optimisation, so I think merging these two like following would be perfect.

I'll also add a test for that case.

```
def _compare_with_regex(request_headers: Union[Dict[Any, Any], Any]) -> bool:
    checked_request_headers = {k: 1 for k in request_headers.keys()}
    for k, v in headers.items():
        if request_headers.get(k) is not None:
            checked_request_headers[k] = 0
            if isinstance(v, re.Pattern):
                if re.match(v, request_headers[k]) is None:
                    return False
            else:
                if not v == request_headers[k]:
                    return False
        elif strict_match:
            return False
    return True if not strict_match else sum(checked_request_headers.values()) == 0
```

geetptl: This will work only when there's a field in `header` which doesn't match anything in `request_header`.

If we have a field in `request_header` that is not present in `header`, this change won't capture that.

Although this is a good optimisation, so I think merging these two like following would be perfect.

```suggestion
        checked_request_headers = {k: 1 for k in request_headers.keys()}
        for k, v in headers.items():
            if request_headers.get(k) is not None:
                checked_request_headers[k] = 0
                if isinstance(v, re.Pattern):
                    if re.match(v, request_headers[k]) is None:
                        return False
                else:
                    if not v == request_headers[k]:
                        return False
            elif strict_match:
                return False
        return True if not strict_match else sum(checked_request_headers.values()) == 0
```

I'll also add a test for that case.

beliaev-maksim: current implementation is not human readable. I suggest we rethink the design to remove counters

geetptl: Fair point, let me add this and all kinda tests for it.

geetptl:
```
def test_request_matches_headers_regex_strict_match():
    @responses.activate
    def run():

        # <some request initialization code here>

        # Subtests
        # 1. Everything matches
        # 2. len(headers) < len(request_headers)
        # 3. len(headers) > len(request_headers)
        # 4. len(headers) = len(request_headers), one field is mismatched
        # 5. len(headers) = len(request_headers), regex mismatch

        test_request_matches_headers_regex_strict_match_1(session, url)

        test_request_matches_headers_regex_strict_match_2(session, url)

        test_request_matches_headers_regex_strict_match_3(session, url)

        test_request_matches_headers_regex_strict_match_4(session, url)

        test_request_matches_headers_regex_strict_match_5(session, url)

        # defs for each function

    run()
    assert_reset()
```

just asking. Is this a good way to write tests? All are of some variation of the same functionality, hence this. Or should I split them all up?

responses/matchers.py
```diff
@@ -402,14 +403,31 @@ def header_matcher(
     :return: (func) matcher
     """
 
+    def _compare_with_regex(request_headers: Union[Dict[Any, Any], Any]) -> bool:
+        if strict_match and len(request_headers) != len(headers):
+            return False
+
+        for k, v in headers.items():
+            if request_headers.get(k) is not None:
```
beliaev-maksim, line 411: what does RFC say about headers. Can a header officially be equal None ?

geetptl:
```
Field values containing CR, LF, or NUL characters are invalid and dangerous ...
```

-- from RFC 9110 [here](https://datatracker.ietf.org/doc/html/rfc9110#name-field-values)

I'd say its safe to assume that any header will not be None

beliaev-maksim: yes, confirm

Headers do not support None and should be at least a whitespace

responses/tests/test_matchers.py
```diff
@@ -782,3 +783,208 @@ def test_matchers_create_key_val_str():
         "a, {3: test, key1: val1, key2: 2}], test: val}"
     )
     assert conv_str == reference
+
+
+def test_request_matches_headers_regex():
```
beliaev-maksim, line 788: can you please create a class for headers matcher and wrap everything into a class. That will allow to logically separate tests

responses/tests/test_matchers.py
```diff
@@ -782,3 +783,166 @@ def test_matchers_create_key_val_str():
         "a, {3: test, key1: val1, key2: 2}], test: val}"
     )
     assert conv_str == reference
+
+
+class TestHeaderWithRegex:
+    url = "http://example.com/"
```
beliaev-maksim, line 789: I recommend to put url into setup method

```python
class A:
    var = 1
    def a(self):
        self.var = 2
        
    def b(self):
        print(self.var)
        
cl = A()
cl.a()
cl.b()
```

beliaev-maksim: that will work as well

jgehrcke:
Nice, thank you!

PR #664, opened 2023-08-17
Calls info in BaseResponse

zeezdev:
Sometimes it is more convenient to get information about the calls made directly from the mockend request, instead of calculating the order of your calls in the global list (`responses.calls[n]`).

What do you think of such a possibility?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1083,47 +1083,69 @@ Assert Request Calls data
 ------------------
 
 ``Request`` object has ``calls`` list which elements correspond to ``Call`` objects
-in the global list of ``Registry``.
+in the global list of ``Registry``. This can be useful when the order of requests is not
+guaranteed, but you need to check their correctness, for example in multithreaded
+applications.
 
 .. code-block:: python
 
+    import concurrent.futures
     import responses
     import requests
 
 
     @responses.activate
     def test_assert_calls_on_resp():
-        rsp = responses.add(responses.GET, "http://www.example.com")
-        rsp2 = responses.put(
-            "http://foo.bar/42/", json={"id": 42, "name": "Bazz"}
+        uid_with_permissions = [
+            ("0123", {"client": True, "admin": False, "disabled": False}),
+            ("1234", {"client": False, "admin": True, "disabled": False}),
+            ("2345", {"client": False, "admin": False, "disabled": True}),
+        ]
+        rsp0123 = responses.patch(
+            "http://www.foo.bar/0123/",
+            json={"OK": True, "uid": "0123"},
+            status=200,
+        )
+        rsp1234 = responses.patch(
+            "http://www.foo.bar/1234/",
+            json={"OK": False, "uid": "1234"},
+            status=400,
+        )
+        rsp2345 = responses.patch(
+            "http://www.foo.bar/2345/",
+            json={"OK": True, "uid": "2345"},
+            status=200,
         )
 
-        requests.get("http://www.example.com")
-        requests.get("http://www.example.com?hello=world")
-        requests.put(
-            "http://foo.bar/42/",
-            json={"name": "Bazz"},
-        )
+        def update_permissions(uid, permissions):
+            url = f"http://www.foo.bar/{uid}/"
+            response = requests.patch(url, json=permissions)
+            return response
+
+        with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
+            future_to_uid = {
+                executor.submit(update_permissions, uid, permissions): uid
+                for (uid, permissions) in uid_with_permissions
+            }
+            for future in concurrent.futures.as_completed(future_to_uid):
+                uid = future_to_uid[future]
+                response = future.result()
+                print("%s updated with %d status code" % (uid, response.status_code))
 
         assert len(responses.calls) == 3  # total calls count
-        # Assert calls to the 1st resource ("http://www.example.com"):
-        assert rsp.call_count == 2
-        assert rsp.calls[0] is responses.calls[0]
-        assert rsp.calls[1] is responses.calls[1]
-        request = rsp.calls[0].request
-        assert request.url == "http://www.example.com/"
-        assert request.method == "GET"
-        request = rsp.calls[1].request
-        assert request.url == "http://www.example.com/?hello=world"
-        assert request.method == "GET"
-        # Assert calls to the 2nd resource (http://foo.bar):
-        assert rsp2.call_count == 1
-        assert rsp2.calls[0] is responses.calls[2]
-        request = rsp2.calls[0].request
-        assert request.url == "http://foo.bar/42/"
-        assert request.method == "PUT"
-        request_payload = json.loads(request.body)
-        assert request_payload == {"name": "Bazz"}
+
+        assert rsp0123.call_count == 1
+        assert rsp0123.calls[0] in responses.calls
+        assert json.loads(rsp0123.calls[0].request.body) == {"client": True, "admin": False, "disabled": False}
+
+        assert rsp1234.call_count == 1
+        assert rsp1234.calls[0] in responses.calls
+        assert json.loads(rsp1234.calls[0].request.body) == {"client": False, "admin": True, "disabled": False}
+        assert rsp1234.calls[0].response.json() == {"OK": False, "uid": "1234"}
+
+        assert rsp2345.call_count == 1
+        assert rsp2345.calls[0] in responses.calls
+        assert json.loads(rsp2345.calls[0].request.body) == {"client": False, "admin": False, "disabled": True}
 
 
 Multiple Responses
--- a/README.rst
+++ b/README.rst
@@ -1096,36 +1096,19 @@ applications.
 
     @responses.activate
     def test_assert_calls_on_resp():
-        uid_with_permissions = [
-            ("0123", {"client": True, "admin": False, "disabled": False}),
-            ("1234", {"client": False, "admin": True, "disabled": False}),
-            ("2345", {"client": False, "admin": False, "disabled": True}),
-        ]
-        rsp0123 = responses.patch(
-            "http://www.foo.bar/0123/",
-            json={"OK": True, "uid": "0123"},
-            status=200,
-        )
-        rsp1234 = responses.patch(
-            "http://www.foo.bar/1234/",
-            json={"OK": False, "uid": "1234"},
-            status=400,
-        )
-        rsp2345 = responses.patch(
-            "http://www.foo.bar/2345/",
-            json={"OK": True, "uid": "2345"},
-            status=200,
-        )
+        rsp1 = responses.patch("http://www.foo.bar/1/", status=200)
+        rsp2 = responses.patch("http://www.foo.bar/2/", status=400)
+        rsp3 = responses.patch("http://www.foo.bar/3/", status=200)
 
-        def update_permissions(uid, permissions):
+        def update_user(uid, is_active):
             url = f"http://www.foo.bar/{uid}/"
-            response = requests.patch(url, json=permissions)
+            response = requests.patch(url, json={"is_active": is_active})
             return response
 
         with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
             future_to_uid = {
-                executor.submit(update_permissions, uid, permissions): uid
-                for (uid, permissions) in uid_with_permissions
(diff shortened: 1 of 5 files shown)


responses/__init__.py
```diff
@@ -1064,14 +1075,16 @@ def _on_request(
                     request, match.get_response(request)
                 )
             except BaseException as response:
-                match.call_count += 1
+                next_index = len(self._calls)
                 self._calls.add(request, response)
+                match.calls.add_call(self._calls[next_index])
```
markstory, line 1080: If you have the request and response, why do you need to read out of `self._calls`? Wouldn't `match.calls.add()` work?

zeezdev: Actually, that's what I did initially. But then an idea came up:
* reuse an already created call from the "global" list `self._calls`
* have the same call instance in the global list and in the individual response list (`match.calls`)

(using `match.calls.add()` we'll create a separate instance of the same call)

markstory: Ok, wanting to reuse the objects seems reasonable to me. What do you think of something like

```python
call = Call(request, response)
self._calls.add_call(call)
match.calls.add_call(call)
```

That would save us having to index/count call objects.

zeezdev: Looks good, thanks, Let me try...

zeezdev: I replaced the index with 

```python
        call = Call(request, response)
        self._calls.add_call(call)
        match.calls.add_call(call)
```

and added an example of usage Response.calls into the README.

README.rst
```diff
@@ -1079,6 +1079,53 @@ Assert that the request was called exactly n times.
...
+    import responses
+    import requests
+
+
+    @responses.activate
+    def test_assert_calls_on_resp():
+        rsp = responses.add(responses.GET, "http://www.example.com")
+        rsp2 = responses.add(
+            responses.PUT, "http://foo.bar/42/", json={"id": 42, "name": "Bazz"}
+        )
```
beliaev-maksim, line 1099:
```suggestion
        rsp2 = responses.put(
            "http://foo.bar/42/", json={"id": 42, "name": "Bazz"}
        )
```

README.rst
```diff
@@ -1079,6 +1079,53 @@ Assert that the request was called exactly n times.
         responses.assert_call_count("http://www.example.com?hello=world", 1) is True
 
 
+Assert Request Calls data
+------------------
+
+``Request`` object has ``calls`` list which elements correspond to ``Call`` objects
+in the global list of ``Registry``.
```
beliaev-maksim, line 1086: Need more context on the issue that it solves for the user rather than the statement of technical fact

README.rst
```diff
@@ -1079,6 +1079,53 @@ Assert that the request was called exactly n times.
...
+        assert request.url == "http://www.example.com/?hello=world"
+        assert request.method == "GET"
+        # Assert calls to the 2nd resource (http://foo.bar):
+        assert rsp2.call_count == 1
+        assert rsp2.calls[0] is responses.calls[2]
+        request = rsp2.calls[0].request
+        assert request.url == "http://foo.bar/42/"
+        assert request.method == "PUT"
+        request_payload = json.loads(request.body)
+        assert request_payload == {"name": "Bazz"}
```
beliaev-maksim, line 1126: this could be sparse to increase readability

beliaev-maksim (changes requested):
also, CHANGELOG entry is missing

zeezdev:
> also, CHANGELOG entry is missing

Added the CHANGELOG entry.

README.rst
zeezdev: I have extended description and replaced the example, please review.

README.rst
zeezdev: The example has been replaced. Is this still relevant?

beliaev-maksim (approved):
LGTM

markstory:
This is looking good, a few small formatting issues in the README that need to be fixed up before this can be merged.

zeezdev:
> This is looking good, a few small formatting issues in the README that need to be fixed up before this can be merged.

Fixed.

zeezdev:
It looks like [mypy problems](https://github.com/getsentry/responses/actions/runs/6585106855/job/17907124070?pr=664#step:5:15) is related to v1.6.1 that [released](https://mypy-lang.blogspot.com/2023/10/mypy-16-released.html) several days ago.

I added several ignore comments related to my changes and old code. Please review again.

responses/matchers.py
```diff
@@ -13,7 +13,7 @@
 from urllib.parse import urlparse
 
 from requests import PreparedRequest
-from requests.packages.urllib3.util.url import parse_url
+from requests.packages.urllib3.util.url import parse_url  # type: ignore[import-untyped]
```
beliaev-maksim, line 16: why is that ?

responses/__init__.py
```diff
@@ -252,11 +258,11 @@ def __len__(self) -> int:
         return len(self._calls)
 
     @overload
-    def __getitem__(self, idx: int) -> Call:
+    def __getitem__(self, idx: int) -> Call:  # type: ignore[misc]
```
beliaev-maksim, line 261: too many ignores, something looks to be incorrect

beliaev-maksim: fixed in https://github.com/getsentry/responses/pull/684

first merge https://github.com/getsentry/responses/pull/684, then current PR and we are good to ship the release

zeezdev: Merged, thanks!

responses/tests/test_responses.py
```diff
@@ -2035,6 +2036,88 @@ def run():
...
+        requests.put(
+            "http://www.foo.bar/42/",
+            json={"name": "Bazz"},
+        )
+
+        assert rsp.call_count == 2
+        request = rsp.calls[0].request
+        assert request.url == "http://www.example.com/"
+        assert request.method == "GET"
+        request = rsp.calls[1].request
```
beliaev-maksim, line 2077: I think we can make it a bit more readable by enumerating `request` objects as well
```suggestion
        assert rsp.call_count == 2
        
        request1 = rsp.calls[0].request
        assert request.url == "http://www.example.com/"
        assert request.method == "GET"
        
        request2 = rsp.calls[1].request
```

and so on

responses/tests/test_responses.py
```diff
@@ -2035,6 +2036,88 @@ def run():
...
+        assert request.url == "http://www.example.com/?hello=world"
+        assert request.method == "GET"
+        assert rsp2.call_count == 1
+        request = rsp2.calls[0].request
+        assert request.url == "http://www.foo.bar/42/"
+        assert request.method == "PUT"
+        request_payload = json.loads(request.body)
```
beliaev-maksim, line 2084: what exactly do we test here ?

why do we need to test that request body could be converted?

responses/tests/test_responses.py
```diff
@@ -2035,6 +2036,88 @@ def run():
...
+            "http://www.foo.bar/42/",
+            json={"name": "Bazz"},
+        )
+
+        assert rsp.call_count == 1
+        response = rsp.calls[0].response
+        assert response.content == b"test"
+        assert response.status_code == 200
+        assert rsp2.call_count == 1
+        response = rsp2.calls[0].response
```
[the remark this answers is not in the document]
zeezdev: agree, done

responses/tests/test_responses.py
zeezdev: (I changed the test a little)

Here we validate that request `rsp2_request1 = rsp2.calls[0].request` obtained from the mock's call contains expected payload `request_payload = {"name": "Bazz"}`, that was used on the put request.

> request body could be converted?

`PreparedRequest` contains only encoded data in the `body` field, so we decode it to compare with the original data - `request_payload`.

beliaev-maksim: > `PreparedRequest` contains only encoded data in the `body` field, so we decode it to compare with the original data - `request_payload`.

correct, but do we need to check it at all as part of `responses` ?

we attach `request` object as is without modification 

https://github.com/getsentry/responses/blob/5c6a3b13bbefc2d4cae0cbd1e9177ab702e15e76/responses/__init__.py#L1041

and in the case of this PR we can skip this check since it does not add a value.

although, you can fire a parallel PR that can add a separate test to validate that we have all the required data is attached to a request
we still will need to discuss if that is required or not but this we can do in a separate thread

README.rst
```diff
@@ -1079,6 +1079,87 @@ Assert that the request was called exactly n times.
...
+
+.. code-block:: python
+
+    import concurrent.futures
+    import responses
+    import requests
+
+
+    @responses.activate
+    def test_assert_calls_on_resp():
```
beliaev-maksim, line 1098: I understand that this example might look very close to what you have in production. But can we unload it a bit and reduce complexity?

we probably still want to show an example of `concurrent` since this could be one of the main usages, but we do not need all those complicated matrices of settings to show the usage

can you please simplify it to remove additional arguments and have more clear names like
rsp1, rsp2, rsp3

beliaev-maksim (changes requested):
I think we are very close to the conclusion

technically all looks good, just need to simplify user facing docs

README.rst
zeezdev: Got you, I'll try to simplify this.

zeezdev: simplified, please take a look

README.rst
```diff
@@ -1079,6 +1079,59 @@ Assert that the request was called exactly n times.
...
+
+        with concurrent.futures.ThreadPoolExecutor(max_workers=3) as executor:
+            future_to_uid = {
+                executor.submit(update_user, uid, is_active): uid
+                for (uid, is_active) in [("3", True), ("2", True), ("1", False)]
+            }
+            for future in concurrent.futures.as_completed(future_to_uid):
+                uid = future_to_uid[future]
+                response = future.result()
+                print("%s updated with %d status code" % (uid, response.status_code))
```
beliaev-maksim, line 1116: Can you please update it with f-string

README.rst
beliaev-maksim: @zeezdev now looks a way better, thank you!

beliaev-maksim (approved):

README.rst
zeezdev: done

beliaev-maksim:
@markstory can you please do a final pass here

responses/tests/test_responses.py
zeezdev: @beliaev-maksim and what about this question?

beliaev-maksim: @zeezdev sorry, missed this one in a long thread. Yes.
And if for some reason we have missed the tests that ensure the quality of `Call` objects. Then I would recommend to create separate tests for those and open it in a separate PR.

responses/tests/test_responses.py
```diff
@@ -2035,6 +2035,36 @@ def run():
...
 
 
+def test_response_calls_and_registry_calls_are_equal():
+    @responses.activate
+    def run():
+        rsp1 = responses.add(responses.GET, "http://www.example.com")
+        rsp2 = responses.add(responses.GET, "http://www.example.com/1")
+        rsp3 = responses.add(
+            responses.GET, "http://www.example.com/2"
+        )  # won't be requested
```
beliaev-maksim, line 2045: that comment is probably true only with the default registry, or?

zeezdev: I mean, this mocked resource (`http://www.example.com/2`) isn't requested during the test to explain why we get empty `rsp3.calls`.
I can remove this if it confuses.

beliaev-maksim: that is fine to keep it for the testing purpose.

I just mean that if the user applies a different registry or we switch the default registry (which is unlikely), then this comment will not be correct

https://github.com/getsentry/responses#ordered-registry

zeezdev: but this comment relates to the specific scope of the test. And is it relevant in this case?

If I were to initialize the test with
`@responses.activate(registry=OrderedRegistry)`

then it would be a different case with different expectations, wouldn’t it?

beliaev-maksim: that is more of a note rather than a call for an action :)

I am +1 with the current state of PR

now wait for @markstory to make his round of review

beliaev-maksim:
@zeezdev great work, thanks!

PR #665, opened 2023-08-20
Support gzipped request.body in json_params_matcher

haakonvt:
I just discovered `matchers` (absolute lifesaver in concurrent tests), but I currently need to disable gzip for the tests since it isn't supported ootb.

Ltmk if there are additional changed/updates/changelog entries etc. you want me to create 😄

### DIFF
--- a/responses/matchers.py
+++ b/responses/matchers.py
@@ -126,11 +126,11 @@ def json_params_matcher(
         request_body = request.body
         json_params = (params or {}) if not isinstance(params, list) else params
         try:
-            if isinstance(request_body, bytes):
+            if isinstance(request.body, bytes):
                 try:
-                    request_body = request_body.decode("utf-8")
+                    request_body = request.body.decode("utf-8")
                 except UnicodeDecodeError:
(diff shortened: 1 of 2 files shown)


doctrino (approved):
Looks good to me

markstory:
Looking good. There is an error from mypy that will need to be resolved before this can be merged.

> responses/matchers.py:132:52: error: Argument 1 to "decompress" has incompatible type "bytes | str"; expected "Buffer"  [arg-type]

haakonvt:
> Looking good. There is an error from mypy that will need to be resolved before this can be merged.
> 
> > responses/matchers.py:132:52: error: Argument 1 to "decompress" has incompatible type "bytes | str"; expected "Buffer"  [arg-type]

That one surprised me, seems like mypy think the error can be raised after variable has been updated... I wasn't sure how you feel about `cast` in this codebase, so I changed a few `request_body` to `request.body` to help `mypy` understand.

markstory:
Thank you 🎉

PR #667, opened 2023-08-24
types-pyyaml is not needed at runtime

asottile-sentry:


beliaev-maksim:
Then people who install the package will have issues in imported module, no?

asottile-sentry:
> Then people who install the package will have issues in imported module, no?

`types-pyyaml` provides only `.pyi` files -- these aren't importable

beliaev-maksim:
Yes, but when they run mypy on their side, all types should be present. Then either users will need to add it to their dependencies or we provide it.

Or does mypy work differently?

asottile-sentry:
mypy will prompt them to install types-pyyaml _if needed_ -- we could add an extra here for mypy but it shouldn't get installed for general users of `responses` that may not be using mypy

beliaev-maksim:
If I remember right, there were complaints in the past about type checks and we added all the deps.
So, it was deliberate choice

@markstory should we stick or change?

I recommend to keep

asottile-sentry:
> If I remember right, there were complaints in the past about type checks and we added all the deps. So, it was deliberate choice
> 
> @markstory should we stick or change?
> 
> I recommend to keep

if that were the case then `types-requests` would also be a hard dependency -- since that also will produce a message from mypy:

```console
$ mypy responses/ | grep requests | head -1
responses/matchers.py:14:1: error: Library stubs not installed for "requests"  [import]
```

but typically typing-only deps are not part of the library deps

markstory:
Users of responses that are using the recorder to create yaml files generally wouldn't have to directly interact with pyyaml. I wasn't able to find any issues asking for type packages to be installed, however I did find one asking for typing packages to not be part of the `install_requires` (#595).

PR #672, opened 2023-09-22
removed support for python3.7

beliaev-maksim:


markstory:
Why though? It isn't causing us any maintenance pain right now.

beliaev-maksim:
@markstory we already did an announcement that the next minor release will not have it and soon we will need to onboard python 3.12

py3.7 is EOL and is not supported by CPy. Users still can use older versions of responses but should not expect maintenance from us

markstory:
> we already did an announcement that the next minor release will not have it

I had forgotten about that, my bad 🤦

PR #676, opened 2023-09-25
Revert "fix mypy issue"

beliaev-maksim:
Reverts getsentry/responses#674

I do not know why, but now it started to work back and CI fails on [master](https://github.com/getsentry/responses/actions/runs/6301576006/job/17106907833) (facepalm)


beliaev-maksim:
what a miracle, both CIs have the same packages installed, how can it be different

@markstory ideas?

markstory:
> ideas?

Nope, this is spooky mypy behavior to me too.

PR #678, opened 2023-10-04
fix CI coverage

beliaev-maksim:
change the order of dependencies installation to ensure that `urllib3` is not overwritten

otherwise we have an issue that we do not test `urllib3<2`


beliaev-maksim:
will merge this fix into #677

Issue #682, opened 2023-10-20
Bug when record the binary response to yaml


narugo1992:
### Describe the bug

When recording the binary response (e.g. image), the response is broken when added from file again.

### Version of `responses`

0.20.0

### Steps to Reproduce

Use this code to record the image's response from danbooru to `danbooru.yaml`

```python
import logging

import requests
from responses import _recorder

logging.basicConfig(level=logging.DEBUG)

IMAGE_URL = 'https://cdn.donmai.us/original/e1/f8/e1f815ed0d489dddda86ac55815ce37f.jpg'


@_recorder.record(file_path='danbooru.yaml')
def main():
    resp = requests.get(IMAGE_URL)
    resp.raise_for_status()


if __name__ == '__main__':
    main()

```

and then use the following code to download this image both from original danbooru site (without mocking) and from mocked responses

```python
import logging
import os

import requests
import responses
from tqdm.auto import tqdm

logging.basicConfig(level=logging.DEBUG)


def download_file(url, filename, expected_size: int = None, desc=None, **kwargs):
    session = requests.session()
    response = session.get(url, stream=True, allow_redirects=True, **kwargs)
    expected_size = expected_size or response.headers.get('Content-Length', None)
    expected_size = int(expected_size) if expected_size is not None else expected_size

    desc = desc or os.path.basename(filename)
    directory = os.path.dirname(filename)
    if directory:
        os.makedirs(directory, exist_ok=True)

    with open(filename, 'wb') as f:
        with tqdm(total=expected_size, unit='B', unit_scale=True, unit_divisor=1024, desc=desc) as pbar:
            for chunk in response.iter_content(chunk_size=1024):
                f.write(chunk)
                pbar.update(len(chunk))

    actual_size = os.path.getsize(filename)
    if expected_size is not None and actual_size != expected_size:
        os.remove(filename)
        raise requests.exceptions.HTTPError(f"Downloaded file is not of expected size, "
                                            f"{expected_size} expected but {actual_size} found.")

    return filename


responses._add_from_file('danbooru.yaml')

IMAGE_URL = 'https://cdn.donmai.us/original/e1/f8/e1f815ed0d489dddda86ac55815ce37f.jpg'


def test_danbooru_origin():
    download_file(IMAGE_URL, 'test_img_actual.jpg')


@responses.activate
def test_danbooru():
    download_file(IMAGE_URL, 'test_img.jpg')


if __name__ == '__main__':
    test_danbooru_origin()
[...]

PR #683, opened 2023-10-20
dev(narugo): add support for headers and binary body in recorder

narugo1992:
related issue: #682

### DIFF
--- a/responses/_recorder.py
+++ b/responses/_recorder.py
@@ -1,3 +1,4 @@
+import base64
 from functools import wraps
 from typing import TYPE_CHECKING
 
@@ -22,6 +23,7 @@ if TYPE_CHECKING:  # pragma: no cover
 
 import yaml
 
+from responses import _UNSET
 from responses import RequestsMock
 from responses import Response
 from responses import _real_send
@@ -45,12 +47,19 @@ def _dump(
     for rsp in registered:
         try:
             content_length = rsp.auto_calculate_content_length  # type: ignore[attr-defined]
+            body = rsp.body  # type: ignore[attr-defined]
+            if isinstance(body, bytes):
+                body = base64.urlsafe_b64encode(body).decode()
+                body_encoded = True
+            else:
+                body_encoded = False
             data["responses"].append(
                 {
                     "response": {
                         "method": rsp.method,
                         "url": rsp.url,
-                        "body": rsp.body,  # type: ignore[attr-defined]
+                        "body": body,  # type: ignore[attr-defined]
+                        "body_encoded": body_encoded,
                         "status": rsp.status,  # type: ignore[attr-defined]
                         "headers": rsp.headers,
                         "content_type": rsp.content_type,
@@ -116,11 +125,18 @@ class Recorder(RequestsMock):
         request.params = self._parse_request_params(request.path_url)  # type: ignore[attr-defined]
         request.req_kwargs = kwargs  # type: ignore[attr-defined]
         requests_response = _real_send(adapter, request, **kwargs)
+        requests_headers = dict(requests_response.headers)
+        if "Content-Type" in requests_headers:
+            requests_content_type = requests_headers.pop("Content-Type")
+        else:
+            requests_content_type = _UNSET
         responses_response = Response(
             method=str(request.method),
             url=str(requests_response.request.url),
             status=requests_response.status_code,
-            body=requests_response.text,
+            headers=requests_headers,
+            body=requests_response.content,
+            content_type=requests_content_type,
         )
         self._registry.add(responses_response)
         return requests_response
--- a/responses/tests/test_recorder.py
+++ b/responses/tests/test_recorder.py
@@ -1,3 +1,4 @@
+import collections.abc
 from pathlib import Path
 
 import pytest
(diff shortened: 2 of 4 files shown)


narugo1992:
one more thing

Aftering saving the binary responses to yaml file (some images from danbooru, 0.5MB~50MB per image) using this pr, I found it be really slow when calling `_all_from_file`, upto 5secs to 5minutes each file.

Maybe a better format should be used to optimize it?

beliaev-maksim:
Performance will be always an issue if we try to save binaries in yaml

We need to enhance the method to allow to supply file path. Then we can store binaries close to yamls. That will be more human readable and performant

Can you please update the PR with the change

narugo1992:
> Performance will be always an issue if we try to save binaries in yaml
> 
> We need to enhance the method to allow to supply file path. Then we can store binaries close to yamls. That will be more human readable and performant
> 
> Can you please update the PR with the change

yep, I am also re-designing these code according to this idea.

draft it for now

narugo1992:
@beliaev-maksim now the binary data are saved to files.

For example, if you specify the config file `responses.yaml`, the binary files will be placed at `responses/1.bin`, `responses/2.bin`, etc. And the relative path of binary files will be put into the `responses.yaml` (the relative paths are relative to `responses.yaml`'s parent directory, so the yaml and binary files can be easily moved to somewhere without any change).

I'm now using the code in this pr, and it works well on ubuntu20.04, Python3.8. Here are some yaml and binary files I created to my unittest (the zip files in this repo): https://huggingface.co/datasets/deepghs/waifuc_responses/tree/main .

I think cross-platform testing is needed now, especially on the Windows platform, and the `/` in the relative path may cause errors.

responses/_recorder.py
```diff
@@ -38,19 +39,44 @@ def _remove_nones(d: "Any") -> "Any":
...
+    dumper_mode: "str" = "w",
+):
     data: Dict[str, Any] = {"responses": []}
+
+    # e.g. config_file = 'my/dir/responses.yaml'
+    # parent_directory = 'my/dir'
+    # binary_directory = 'my/dir/responses'
+    fname, fext = os.path.splitext(os.path.basename(config_file))
+    parent_directory = os.path.dirname(os.path.abspath(config_file))
+    binary_directory = os.path.join(
```
beliaev-maksim, line 53: let's use `pathlib`

responses/tests/test_recorder.py
```diff
@@ -15,15 +17,126 @@
...
+            if not isinstance(other, list):
+                return False
+            if len(self.obj) != len(other):
+                return False
+            for i, (obj_item, other_item) in enumerate(zip(self.obj, other)):
+                if _CompareDict(obj_item) != other_item:
+                    return False
+            return True
+        else:
+            return self.obj == other
```
beliaev-maksim, line 49: why do we need all of those?

responses/_recorder.py
```diff
@@ -38,19 +39,44 @@ def _remove_nones(d: "Any") -> "Any":
 
 def _dump(
     registered: "List[BaseResponse]",
-    destination: "Union[BinaryIO, TextIOWrapper]",
+    config_file: "Union[str, os.PathLike]",
```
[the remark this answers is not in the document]
beliaev-maksim: that I understand. Why is this change necessary ?

responses/tests/test_recorder.py
narugo1992: Because the unittest need to compare the full dict data.

In the new version, headers are added into the recorded data, which may contain something like `Python/3.x.y` or something else which are not important for unittesting. So the `_CompareDict` and `_NOT_CARE` object are added to ignore these values.

responses/_recorder.py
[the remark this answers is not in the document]
beliaev-maksim: the question is not about the naming of rge arguments but more about why we need to change the existing behaviour ?

what is the benefit of doing so over the initial implementation ?

example_bins/202.bin
```diff
@@ -0,0 +1 @@
+OK
```
beliaev-maksim, line 1:
```suggestion
OK

```

example_bins/400.bin
```diff
@@ -0,0 +1 @@
+Invalid status code
```
beliaev-maksim, line 1:
```suggestion
Invalid status code

```

example_bins/500.bin
```diff
@@ -0,0 +1 @@
+500 Internal Server Error
```
beliaev-maksim, line 1:
```suggestion
500 Internal Server Error

```

beliaev-maksim (changes requested):

example_bins/500.bin
narugo1992: The same as above.

example_bins/400.bin
narugo1992: The same as above.

responses/_recorder.py
narugo1992: Changed.

responses/_recorder.py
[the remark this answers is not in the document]
narugo1992: Because in the new version, we need to save the response body to `.bin` files, which means we need to write other files besides `responses.yaml`. If you still maintain the original mode of passing in the output buffer, it is obviously impossible to do this.

narugo1992:
Discovered a new issue: 'responses' cannot handle HEAD requests with the `Content-Length` header properly, causing an `IncompleteRead` exception to be thrown during length checking in urllib3.

narugo1992:
@beliaev-maksim @markstory 

Now ready for a review, as this version has already been deployed in our own project and is running as expected. 

Additionally, all unit tests have passed in Github Actions.

christianfosli:
@narugo1992 Did you by any chance publish your fork anywhere?

I'm using responses for testing with an external API which recently switched one of their endpoints from xml text to zip archive. Would be great to try this out, hopefully without needing to learn how to build it myself :sweat_smile:

PR #684, opened 2023-10-20
fix types

beliaev-maksim:
* remove dependency on urllib3 vendored in requests since it is not maintained anymore. See https://github.com/python/typeshed/issues/6893#issuecomment-1012511758
* new version of mypy does not require overloads for type checking. `mypy` errors and complains that `overload`  itself has `Any` argument (which is obvious). In theory we can suppress the error with the comment, but at the same time we can just remove overloads as not required. Do not have strong opinion on this.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -51,7 +51,7 @@ from urllib.parse import urlunsplit
 
 from urllib3.response import HTTPHeaderDict
 from urllib3.response import HTTPResponse
-from urllib3.util.url import parse_url  # pragma: no cover
+from urllib3.util.url import parse_url
 
 if TYPE_CHECKING:  # pragma: no cover
     # import only for linter run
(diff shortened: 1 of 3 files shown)


responses/__init__.py
```diff
@@ -962,23 +942,6 @@ def __exit__(self, type: Any, value: Any, traceback: Any) -> bool:
...
-        *,
-        registry: Type[Any] = ...,
-        assert_all_requests_are_fired: bool = ...,
-    ) -> Callable[["_F"], "_F"]:
-        """Overload for scenario when
-        'responses.activate(registry=, assert_all_requests_are_fired=True)' is used.
-
-        See https://github.com/getsentry/responses/pull/469 for more details
-        """
-
```
andersk, line 981: These were important too.

- #692

PR #690, opened 2023-11-05
Fix/calllist typing

howamith:
Typing for the `CallList` class was improved in 0.23.0 (see #593), however this has regressed in 0.24.0 - I believe through the changes made in #684. This PR reintroduces the improved typing from #593.

I've tried to add a test to ensure this doesn't happen again, however it's currently evergreen as `mypy` isn't run on the tests. (IMHO `mypy` - and any other static analysis tools - _should_ be run against the tests. Type hints are great and their use should be encouraged, however by not checking the tests there's nothing checking how these types work from the perspective of the code consuming this package - which is probably how this was missed in #684.)

I did have a brief go at running `mypy` over the tests to give the new test a bit more purpose (and to improve coverage generally), but there were 754 errors so that feels like something that should be addressed in its own PR.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -3,7 +3,6 @@ import inspect
 import json as json_module
 import logging
 import socket
-from collections import namedtuple
 from functools import partialmethod
 from functools import wraps
 from http import client
@@ -18,6 +17,7 @@ from typing import Iterable
 from typing import Iterator
 from typing import List
 from typing import Mapping
+from typing import NamedTuple
 from typing import Optional
 from typing import Sequence
 from typing import Sized
@@ -97,7 +97,11 @@ if TYPE_CHECKING:  # pragma: no cover
     ]
 
 
-Call = namedtuple("Call", ["request", "response"])
+class Call(NamedTuple):
+    request: "PreparedRequest"
+    response: "_Body"
+
+
 _real_send = HTTPAdapter.send
 _UNSET = object()
 
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -24,6 +24,7 @@ from typing import Sized
 from typing import Tuple
 from typing import Type
 from typing import Union
+from typing import overload
 from warnings import warn
 
 import yaml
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -61,7 +61,7 @@ def run_tests(self):
 
 setup(
     name="responses",
-    version="0.24.0",
```
beliaev-maksim, line 64: We don't change the version in setup
Please revert

responses/__init__.py
```diff
@@ -241,6 +246,14 @@ def __iter__(self) -> Iterator[Call]:
     def __len__(self) -> int:
         return len(self._calls)
 
+    @overload
+    def __getitem__(self, idx: int) -> Call:
```
beliaev-maksim, line 250: Recent version of mypy doesn't require overload, why do we necessarily need it?

beliaev-maksim (changes requested):

responses/__init__.py
stinovlas: I have `mypy==1.6.1`, which is the most current version and it still throws:

```
error: Item "list[Call]" of "Call | list[Call]" has no attribute "request"  [union-attr]
```

Changes in this PR should fix that. I can't really imagine how would `mypy` guess the appropriate combination of input and output types without overload.

beliaev-maksim: The previous CI passed just fine with mypy, although I saw some flakiness in results in the last couple of months

What is the difference then?

howamith: That will be because `mypy` isn't being run against the tests. The definition of `CallList`'s `__getitem__` method is fine. It's where it's used that becomes a problem. You can reproduce this locally removing the line that excludes the tests from `mypy` in the .ini file.

This is why I've suggested running static analysis on the tests.

beliaev-maksim: then we need to indeed enabling mypy on tests to validate this PR and all the future PRs

howamith: Is it worth getting this out to fix the regression, and then dealing with enabling static analysis in the tests in a separate PR? My original intention was to do that here but after removing the exclusion from the mypy config file there were *754* errors. Fixing those is going to make the diff really noisy and feels like a separate issue (IMO).

PR #702, opened 2024-01-26
fix: Missing header detection for the non-strict case

woodyza:
Hi!

When #651 was implemented the header matching rule was subtly broken, making it no-longer detect when the specified headers were missing entirely (vs present, but with the wrong value) unless you used strict matching. 

As far as I understand that is not the intention, and you should be able to use the header matcher to detect missing headers without necessarily caring about all of the other headers in the request. It's a small change to enable that, and it preserves the strict match and regex match logic.

I've added a test case that fails without the current change, and the rest of the cases still pass with it in place.


markstory:
Thank you 🎉

woodyza:
Thanks @markstory for the merge 🙏 Any idea when `0.25.0` might come out?

markstory:
@woodyza I started a release build just now.

woodyza:
Awesome, thanks!

Issue #708, opened 2024-03-14
With `urllib3>=2`, mocking HEAD response with content-length header set requires body


KennyChenBasis:
### Describe the bug

With `urllib3>=2`, mocking HEAD responses with content-length header set requires a body with the right length, otherwise an `IncompleteRead` is raised. This should not be the case since HEAD responses do not have bodies.

### Additional context

The fix is probably to pass the request method to `HTTPResponse` when forming the response. See the code around https://github.com/urllib3/urllib3/blob/733f638a2faa02b4ff8a9f3b5668949d39396b8b/src/urllib3/response.py#L624.

### Version of `responses`

0.25.0

### Steps to Reproduce

Run `pytest` on 
```python
import requests
import responses


def test_download():
    url = 'http://test'
    content = b'download me'
    headers = {'Content-length': str(len(content))}
    with responses.RequestsMock() as rsps:
        rsps.head(url, headers=headers)

        res = requests.head(url)
        res.raise_for_status()
        assert int(res.headers['content-length']) == 11
        assert res.text == ''
```

### Expected Result

Test passes, like with `urllib3<2`.

### Actual Result

Test fails with `requests.exceptions.ChunkedEncodingError: ('Connection broken: IncompleteRead(0 bytes read, 11 more expected)', IncompleteRead(0 bytes read, 11 more expected))`.

PR #712, opened 2024-05-03
fix mocked HEAD response when content-length header is present

ddarricau:
This PR is a fix for this issue https://github.com/getsentry/responses/issues/708.
Basically when mocking a response to a HEAD request (which contains a content-length header), we do not pass the request_method. Which results in wrongfully raising a `ChunkedEncodingError`


markstory:
Thank you 🎉

PR #717, opened 2024-05-18
Add feature for allowing the body of a request to be matched Closes #716

### LINKED ISSUE #716: Add capability for matching non json request bodies
At work I use responses for mocking api responses for the applications I develop. One of the APIs my application is interacting with does not take input in json format.

A call to this api looks like this

'''Curl api.company.com/v1/id --data '12345' '''

I would like to create a matcher for non json in the body.

I am pretty new to open source and only have some bug fixes and documentation fixes so I would like to contribute this feature back to responses. I think I can code this up from looking at the source. But would responses accept this feature?

TheJacobWalters:
I use responses to mock APIs for applications. One of the API's used by my organization is an API for getting identifying information about about the "Owners" of AWS accounts. And the API just takes in the account number as the body of a GET request. 

here is an example of how I query the API from the command line
```
curl api.company.com/v1/accounts --data 12345
```

I have not been able to use responses to mock this API because this data is not JSON. I would like to contribute back a more flexible matcher for the bodies that just lets you match anything sent to an API endpoint.

### DIFF
--- a/responses/matchers.py
+++ b/responses/matchers.py
@@ -31,11 +31,26 @@ def _filter_dict_recursively(
     return filtered_dict
 
 
+def body_matcher(params: str, *, allow_blank: bool = False) -> Callable[..., Any]:
+    def match(request: PreparedRequest) -> Tuple[bool, str]:
+        reason = ""
+        if isinstance(request.body, bytes):
+            request_body = request.body.decode("utf-8")
+        else:
+            request_body = request.body
+        valid = True if request_body == params else False
+        if not valid:
+            reason = f"request.body doesn't match {params} doesn't match {request_body}"
+        return valid, reason
+
+    return match
+
+
 def urlencoded_params_matcher(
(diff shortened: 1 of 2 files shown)


responses/matchers.py
```diff
@@ -159,7 +174,8 @@ def query_param_matcher(
         conjunction with ``strict_match=False``.
     strict_match : bool, default=True
         If set to ``True``, validates that all parameters match.
-        If set to ``False``, original request may contain additional parameters.
+        If set to ``False``, original request may contain additional parameters.request_body
+        = request.bodyrequest_body = request.body.decode("utf-8").decode("utf-8")
```
markstory, line 178: This looks like a mistake.

TheJacobWalters: Yes that was, I have removed it now

markstory:
Thank you 🎉

PR #720, opened 2024-05-25
Addressed bug where recording requests does not save headers to file. Closes #715

### LINKED ISSUE #715: Cannot record header to file
### Describe the bug

When recording a request that contains a location header, that header is not there anymore, when replaying that request.

### Additional context

It seems the code in [_add_from_file()](https://github.com/getsentry/responses/blob/83f2e287768de4bdd37ec1c144f56aafe38e29de/responses/__init__.py#L850) doesn't seem to restore headers, while the code in [_dump()](https://github.com/getsentry/responses/blob/83f2e287768de4bdd37ec1c144f56aafe38e29de/responses/_recorder.py#L50) does seem to save them.

### Version of `responses`

0.25.0

### Steps to Reproduce

```python
>>> import responses
>>> responses.post(url="fnord", status=201, headers=dict(Location='fnord'))
<Response(url='fnord' status=201 content_type='text/plain' headers='{"Location": "fnord"}')>
>>> responses.registered()
[<Response(url='fnord' status=201 content_type='text/plain' headers='{"Location": "fnord"}')>, <Response(url='fnord' status=201 content_type='text/plain' headers='{"Location": "fnord"}')>]
>>> import responses._recorder
>>> import yaml
>>> responses._recorder._dump(responses.registered(), open('/dev/stdout', 'w'), yaml.dump)
responses:
- response:
    body: ''
    content_type: text/plain
    headers:
      Location: fnord
    method: POST
    status: 201
    url: fnord
>>> responses._recorder._dump(responses.registered(), open('/tmp/dump', 'w'), yaml.dump)
>>> responses._add_from_file(file_path="/tmp/dump")
>>> responses.registered()
[<Response(url='fnord' status=201 content_type='text/plain' headers='{"Location": "fnord"}')>, <Response(url='fnord' status=201 content_type='text/plain' headers='null')>]
```

### Expected Result

Interestingly this does save the location headers, while they are not saved in my application where I discovered this. Still when loading the dump, the headers are ignored.

I'd expect the location header to survive the save / restore.

### Actual Result

Location header gets stripped by the replay from file.

TheJacobWalters:
I have made a working solution here.

I would welcome feedback on how I can improve this. 

I left some comments in my code explaining the main issue I was overcoming. Which is that the headers field of the Request is a unique data structure called ValueView. Writing this data structure to a YAML file created a lot of strange "Tag" artifacts in the file. Therefore I decided that I would need to coerce the data structure into a standard type of Dict. That was harder than I expected and I ended up just parsing the string representation of the ValueView into a Dict. There is probably a better way to do this but I don't know what it is yet.

When this MR is ready to move off of Draft, I can remove those comments.


responses/_recorder.py
```diff
@@ -116,11 +139,28 @@ def _on_request(
...
+        # headers is requests_response.values() returns a ValueView
+        # converting that to a dictionary with the dict() method
+        # resulted in *** ValueError: dictionary update sequence
+        # element #0 has length 28; 2 is required
+        # I need to coerce the key:value maping of headers into a dictionary by some method
+        # If I dont and try and write the key:value to the recording file,
+        # it will have pyyaml tags on them which can't be safe loaded back in
+        # like this
+        # The best way I found was to parse it as a string, and then use
+        # ast to evaluate it but I know this is hacky.
```
markstory, line 151: I'm not comfortable relying on AST here. Couldn't we use a dict comprehension to clone headers into a new dictionary?

```python
header_values = {key: value for key, value in requests_response.headers.items()}
```

TheJacobWalters: Ok, thank you for that recommendation of a dictionary comprehension. I have incorporated that.

responses/_recorder.py
```diff
@@ -36,6 +36,28 @@ def _remove_nones(d: "Any") -> "Any":
...
 
 
+def _remove_default_headers(data: "Any") -> "Any":
+    """
+    It would be too verbose to store these headers in the file generated by the
+    record functionality.
+    """
+    if isinstance(data, dict):
+        keys_to_remove = [
+            "Content-Length",
```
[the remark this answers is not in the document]
TheJacobWalters: Ok that header has been added!

PR #725, opened 2024-06-28
Add support for drop-in replacement Niquests when installed

### LINKED ISSUE #710: Compatibility with Niquests
Hello,

I am writing to see if there would be room for a potential improvement. It is also an opportunity to showcase this http client to Sentry folks.

"responses" is highly bound to Requests and I wondered if there was a possibility to allow a "drop-in" compatibility with [Niquests](https://github.com/jawah/niquests).

Today, the only way to make it work outside of Requests [is to tamper with sys.module](https://niquests.readthedocs.io/en/latest/dev/migrate.html#maintainer-migration), which is kind of unreasonable.

I have several option that can be considered without any default behavior breakage

A) Precedence

```python
try:
    from niquests.adapters import HTTPAdapter
    from niquests.adapters import MaxRetryError
    from niquests.exceptions import ConnectionError
    from niquests.exceptions import RetryError
except ImportError:
    from requests.adapters import HTTPAdapter
    from requests.adapters import MaxRetryError
    from requests.exceptions import ConnectionError
    from requests.exceptions import RetryError
```

B) Environment

```sh
RESPONSES_BACKEND_NIQUESTS=1 pytest tests/
```

C) Expose kwarg in annotations

```python
@responses.activate(for_niquests=True)
def ...
```

Of course if none suit you, feel free to close it.
I am willing to propose a PR.

Regards,

Ousret:
Close #710 
Close https://github.com/jawah/niquests/issues/109

responses remain attached to Requests indefinitely, but there is an opportunity to support another http client fully compatible with Requests, namely Niquests.

The commit add a thin layer of compat logic to import Niquests instead of Requests if present in the environment. At the time of this commit, Niquests is not likely to be present in one's environment unless that is what the user wanted.

An escape hatch is also present, in case both Niquests and Requests are present and the user wants to mock Requests only. Setting "RESPONSES_USE_REQUESTS" with any value inside revert to using Requests only.

Unfortunately, for the time being, no simple ways emerge to support both library at once.

Comments will be attached to justify every changes. I tried to make it as light as possible.
Of course, if anything started to make your life harder, I will surely come and help as much as I can.

This would offer Niquests users, a reliable way to mock http responses.
Lastly, Niquests does have async via `AsyncSession`, that is a mirror of its synchronous counterpart, and for now, I have no idea how it would be conceivable to support it here. It's not the topic of this PR. Just to brought it up.

regards,

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -26,15 +26,17 @@ from typing import overload
 from warnings import warn
 
 import yaml
-from requests.adapters import HTTPAdapter
-from requests.adapters import MaxRetryError
-from requests.exceptions import ConnectionError
-from requests.exceptions import RetryError
 
 from responses.matchers import json_params_matcher as _json_params_matcher
 from responses.matchers import query_string_matcher as _query_string_matcher
 from responses.matchers import urlencoded_params_matcher as _urlencoded_params_matcher
 from responses.registries import FirstMatchRegistry
+
+from ._compat import MOCKED_LIBRARY
+from ._compat import ConnectionError
+from ._compat import HTTPAdapter
+from ._compat import MaxRetryError
+from ._compat import RetryError
 
 try:
     from typing_extensions import Literal
@@ -469,7 +471,9 @@ class BaseResponse:
             if _has_unicode(url):
                 url = _clean_unicode(url)
 
-            return _get_url_and_path(url) == _get_url_and_path(other)
+            return _get_url_and_path(url) == _get_url_and_path(
+                _ensure_url_default_path(other)  # type: ignore[arg-type]
+            )
 
         elif isinstance(url, Pattern) and url.match(other):
             return True
@@ -721,7 +725,7 @@ class RequestsMock:
         assert_all_requests_are_fired: bool = True,
         response_callback: Optional[Callable[[Any], Any]] = None,
         passthru_prefixes: Tuple[str, ...] = (),
-        target: str = "requests.adapters.HTTPAdapter.send",
+        target: str = f"{MOCKED_LIBRARY}.adapters.HTTPAdapter.send",
         registry: Type[FirstMatchRegistry] = FirstMatchRegistry,
         *,
         real_adapter_send: "_HTTPAdapterSend" = _real_send,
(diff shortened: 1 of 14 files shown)


responses/__init__.py
```diff
@@ -469,7 +471,9 @@ def _url_matches(self, url: "_URLPatternType", other: str) -> bool:
             if _has_unicode(url):
                 url = _clean_unicode(url)
 
-            return _get_url_and_path(url) == _get_url_and_path(other)
+            return _get_url_and_path(url) == _get_url_and_path(
+                _ensure_url_default_path(other)  # type: ignore[arg-type]
```
Ousret, line 475: To explain this change:

Requests, in "prepare_url" does things that end to "enforce" a trailing "/". Which isn't enforced by Niquests.

So, we have to enforce consistency across both library. here, only Niquests is affected. Not Requests.

responses/tests/test_responses.py
```diff
@@ -822,8 +844,11 @@ def send(
             verify=True,
             cert=None,
             proxies=None,
+            **kwargs,
```
Ousret, line 847: Niquests adapters have a few more arguments. So this is mainly for the compat purposes. By my reading, it does not affect your initial test.

responses/tests/test_responses.py
```diff
@@ -1058,7 +1094,7 @@ def run():
             body="ok",
             stream=responses_stream,
         )
-        session = requests.session()
+        session = requests.Session()
```
Ousret, line 1097: the `session` with `s` lowercased is actually deprecated since a long time in Requests (since v1 if I recall correctly).
and removed in Niquests.

tox.ini
```diff
@@ -5,6 +5,7 @@ envlist = py38,py39,py310,py311,py312,mypy,precom
 filterwarnings =
     error
     default::DeprecationWarning
+    ignore:Passing msg:DeprecationWarning
```
Ousret, line 8: this deprecation happen in Niquests low-level replacement of urllib3. Nothing to be concerned about.

responses/__init__.py
```diff
@@ -26,16 +26,18 @@
...
-from requests.adapters import MaxRetryError
-from requests.exceptions import ConnectionError
-from requests.exceptions import RetryError
 
 from responses.matchers import json_params_matcher as _json_params_matcher
 from responses.matchers import query_string_matcher as _query_string_matcher
 from responses.matchers import urlencoded_params_matcher as _urlencoded_params_matcher
 from responses.registries import FirstMatchRegistry
 
+from ._compat import MOCKED_LIBRARY
```
Ousret, line 35: After thinking about it, I did not find a perfect way to support Niquests "perfectly" without doing the compat import switch.
I did found that this way offer more stability.

Issue #724, opened 2024-06-24
_recorder.record captures header 'content-encoding: gzip' but does not return gzip content on replay


chriskauffman37:
### Describe the bug

`headers` has been added to the recording file which may include `content-encoding`. When a recording with `content-encoding: gzip` is replayed an error is received:

> urllib3.exceptions.DecodeError: ('Received response with content-encoding: gzip, but failed to decode it.', error('Error -3 while decompressing data: incorrect header check'))

### Version of `responses`

0.25.3

### Steps to Reproduce

```python
# Recording code

@_recorder.record(file_path="out.yaml")
def record():
    # call lookup API
```

```yaml
- response:
    auto_calculate_content_length: false
    body: '[{"first_name":true,"last_name":true,"email":false,"phone_number":false,"custom_id":true,"user_id":"abc123","status":{"code":"ACTIVE","description":"The
      user is fully active with all features enabled. Further restrictions may apply
      for some features."},"employers":[{"id":"abc123","sso_required":false}]}]'
    content_type: text/plain
    headers:
      API-Version: '1'
      X-Dep-Next: T-0.8.4
      X-Download-Options: noopen
      X-Permitted-Cross-Domain-Policies: none
      X-Request-Id: f8903c30-4fb2-4b30-95b1-b20e0fdfb515
      X-Runtime: '0.129368'
      apigw-requestid: Z0iA3ibpIAMEaxg=
      cache-control: max-age=0, private, must-revalidate
      content-encoding: gzip
      etag: W/"2b965752b6adbb3fd7467a489ad314b0"
      referrer-policy: strict-origin-when-cross-origin
      vary: Accept-Encoding, Origin
      x-content-type-options: nosniff
      x-frame-options: SAMEORIGIN
      x-xss-protection: '0'
    method: GET
    status: 200
    url: https://api.com/v1/users/lookup?custom_id=060419806551&last_name=Murphy&first_name=Lucy
```

```python
# Test code

@responses.activate
def test_lookup_api():
    responses._add_from_file(file_path="out.yaml")

    lookup = # call lookup API
    assert lookup["first_name"] == True
```

### Expected Result

Pass

### Actual Result

urllib3.exceptions.DecodeError: ('Received response with content-encoding: gzip, but failed to decode it.', error('Error -3 while decompressing data: incorrect header check'))

markstory:
Storing gzip data in yaml files feels gross. Removing the `content-encoding` header and storing the response body as str/bytes seems more human-friendly to me.

PR #728, opened 2024-07-10
recorder: use a case-insensitive HTTP headers check

ptalbert:
_recorder._remove_default_headers() does a case-sensitive check for HTTP header field names are case insensitive. The result is that for example, a lower-case 'content-type' field will not be removed. Later, when _add_from_file tries to load the data using RequestsMock.add(), a RuntimeError exception will be raised when it finds the 'content_type' kwarg was passed and 'content-type' is in the 'headers' kwargs.

Fix this by simply having _remove_default_headers() do a case-insensitive check.

### DIFF
--- a/responses/_recorder.py
+++ b/responses/_recorder.py
@@ -144,13 +144,20 @@ class Recorder(RequestsMock):
         headers_values = {
             key: value for key, value in requests_response.headers.items()
         }
-        responses_response = Response(
-            method=str(request.method),
-            url=str(requests_response.request.url),
-            status=requests_response.status_code,
-            body=requests_response.text,
-            headers=headers_values,
-        )
+
+        response_params = {
+            "method": str(request.method),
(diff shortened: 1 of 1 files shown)


responses/_recorder.py
```diff
@@ -143,13 +144,20 @@ def _on_request(
...
+            "method": str(request.method),
+            "url": str(requests_response.request.url),
+            "status": requests_response.status_code,
+            "body": requests_response.text,
+            "headers": headers_values,
+        }
+
+        # If the header has a content type then pass it in.
+        if content_type := requests_response.headers.get("content-type"):
+            response_params["content_type"] = content_type
```
markstory, line 158: It would be good to have a test covering this change and validating that the Response that is created from the recording has the right parameters/headers.

andrew-womeldorf-smar:
This should also resolve #724

PR #731, opened 2024-07-16
Fix use of global assert_all_requests_are_fired. Closes #726

### LINKED ISSUE #726: Inconsistent behaviour of assert_all_requests_are_fired between decorator and context manager
### Describe the bug

When a `RequestsMock()` is used as a context manager, `assert_all_requests_are_fired` is defaulted to True and an exception is raised at the end of the context block if a response is unused. However, using the `@responses.activate` decorator on a test function defaults `assert_all_requests_are_fired` to False leading to inconsistent behaviour.

### Additional context

As a workaround, I have tried setting `responses.mock.assert_all_requests_are_fired = True` in conftest.py or just after an import of requests. However, this does not change the behaviour. The issue appears to be that the value in the default mock is never used as it gets defaulted by a default parameter in `activate()`, see: [code](https://github.com/getsentry/responses/blob/f7a7945991efc7cb8133b9d6ec75761b459fe75c/responses/__init__.py#L1017).

### Version of `responses`

0.25.3

### Steps to Reproduce

```python
import responses
import requests


def test_my_api():
        rsps.add(
            responses.GET,
            "http://twitter.com/api/1/foobar",
            body="{}",
            status=200,
            content_type="application/json",
        )


@responses.activate
def test_my_api_2():
    responses.add(
        responses.GET,
        "http://twitter.com/api/1/foobar",
        body="{}",
        status=200,
        content_type="application/json",
    )
```

### Expected Result

Both test cases to fail with:
```
AssertionError: Not all requests have been executed [('GET', 'http://twitter.com/api/1/foobar')]
```

### Actual Result

First test case fails.
Second test case passes.

andrew-cybsafe:
A somewhat naive fix for the global assert_all_requests_are_fired. Tests are passing locally on 3.12.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -202,6 +202,8 @@ def get_wrapped(
         Wrapped function
 
     """
+    if assert_all_requests_are_fired is None:
+        assert_all_requests_are_fired = mock.assert_all_requests_are_fired
     assert_mock = std_mock.patch.object(
         target=responses,
         attribute="assert_all_requests_are_fired",
@@ -1014,7 +1016,7 @@ class RequestsMock:
         func: Optional["_F"] = None,
         *,
         registry: Optional[Type[Any]] = None,
-        assert_all_requests_are_fired: bool = False,
+        assert_all_requests_are_fired: bool | None = None,
     ) -> Union[Callable[["_F"], "_F"], "_F"]:
         if func is not None:
             return get_wrapped(func, self)
--- a/responses/tests/test_responses.py
+++ b/responses/tests/test_responses.py
@@ -1192,6 +1192,25 @@ def test_assert_all_requests_fired_multiple():
     assert_reset()
 
 
+def test_assert_all_requests_fired_global():
+    try:
+        old_value = responses.mock.assert_all_requests_are_fired
+        responses.mock.assert_all_requests_are_fired = True
+
+        @responses.activate
+        def test_some_function():
+            # Not all mocks are called so we'll get an AssertionError
+            responses.add(responses.GET, "http://other_url", json={})
+            responses.add(responses.GET, "http://some_api", json={})
+            requests.get("http://some_api")
+
+        with pytest.raises(AssertionError):
+            test_some_function()
+        assert_reset()
+    finally:
+        responses.mock.assert_all_requests_are_fired = old_value
+
+
 def test_allow_redirects_samehost():
     redirecting_url = "http://example.com"
     final_url_path = "/1"


responses/__init__.py
```diff
@@ -1014,7 +1016,7 @@ def activate(
         func: Optional["_F"] = None,
         *,
         registry: Optional[Type[Any]] = None,
-        assert_all_requests_are_fired: bool = False,
+        assert_all_requests_are_fired: bool | None = None,
```
markstory, line 1019:
```suggestion
        assert_all_requests_are_fired: Optional[bool] = None,
```

We need to use `Optional` for compatibility with older pythons.

andrew-cybsafe:
@markstory can you kick off the tests again?

andrew-cybsafe:
> I think we decided to deprecate this completely a while ago

@beliaev-maksim the linked doc deprecates `responses.assert_all_requests_are_fired` for `responses.mock.assert_all_requests_are_fired`. It's the latter / non-deprecated option that isn't working correctly.

responses/__init__.py
```diff
@@ -202,6 +202,8 @@ def get_wrapped(
         Wrapped function
 
     """
+    if assert_all_requests_are_fired is None:
+        assert_all_requests_are_fired = mock.assert_all_requests_are_fired
```
beliaev-maksim, line 206: that makes the method dependent on the global variable `mock`. If someone downstream relies on this method, then it will get broken.

will that work?
```suggestion
        assert_all_requests_are_fired = responses.assert_all_requests_are_fired
```

beliaev-maksim (changes requested):
it feels like the logic should be moved down to the `activate()` method. 
Have you exhausted all the options to make the change self contained within `activate` ?

responses/__init__.py
andrew-cybsafe: Changed.

andrew-cybsafe:
> it feels like the logic should be moved down to the `activate()` method. Have you exhausted all the options to make the change self contained within `activate` ?

@beliaev-maksim 
The attribute `assert_all_requests_are_fired` is mocked within the get_wrapped() function regardless of whether a value has been passed into the function. This seems to make it difficult to fix this issue outside of get_wrapped(). I did try only mocking the attribute if the value was passed in, e.g. with this code:
```
    if assert_all_requests_are_fired is not None:
        assert_mock = std_mock.patch.object(
            target=responses,
            attribute="assert_all_requests_are_fired",
            new=assert_all_requests_are_fired,
        )
    else:
        from contextlib import nullcontext
        assert_mock = nullcontext()
```

This mostly works except for the test_nested_decorators test, but given I only have a superficial understanding of this code, delving into that is more than I can take on right now.

PR #733, opened 2024-07-23
ci: dependency review action

mdtro:
Add dependency review action. Any new dependencies with known high or critical vulnerabilities will fail CI.


mdtro:
Will use global workflow and ruleset.

PR #736, opened 2024-08-31
Read data from file-like objects in requests (#719)

### LINKED ISSUE #719: Cannot match request with `data=filehandle`
### Describe the bug

When mocking responses to a request where an open filehandle is passed as the `data` argument to the request function/method, the prepared request passed to the matcher has its `body` attribute set to the filehandle rather than to the filehandle's contents (as would be sent in an actual request).

### Additional context

_No response_

### Version of `responses`

0.25.0

### Steps to Reproduce

Run `pytest` on the following file:

```python
from __future__ import annotations
from pathlib import Path
import requests
import responses


@responses.activate
def test_post_file(tmp_path: Path) -> None:
    CONTENT = b"This is test text.\n"

    def match_body(req: requests.PreparedRequest) -> tuple[bool, str]:
        if req.body == CONTENT:
            return (True, "")
        else:
            return (False, f"Request body is not the expected content: {req.body!r}")

    responses.post(
        "http://example.nil/endpoint",
        status=200,
        json={"success": True},
        match=[match_body],
    )
    p = tmp_path / "foo.txt"
    p.write_bytes(CONTENT)
    with requests.Session() as s:
        with p.open("rb") as fp:
            assert s.post("http://example.nil/endpoint", data=fp) == {"success": True}
```

### Expected Result

Matcher should successfully match the request body against the expected bytes, resulting in the test passing without error.

### Actual Result

```
============================= test session starts ==============================
platform darwin -- Python 3.12.3, pytest-8.2.1, pluggy-1.5.0
rootdir: /Users/jwodder/work/dev/tmp/responses-bug
collected 1 item

test.py F                                                                [100%]

=================================== FAILURES ===================================
________________________________ test_post_file ________________________________

tmp_path = PosixPath('/private/var/folders/l7/wrkq93d133d8zpn36fmqrq0r0000gn/T/pytest-of-jwodder/pytest-91/test_post_file0')

    @responses.activate
    def test_post_file(tmp_path: Path) -> None:
        CONTENT = b"This is test text.\n"
    
        def match_body(req: requests.PreparedRequest) -> tuple[bool, str]:
            if req.body == CONTENT:
                return (True, "")
            else:
                return (False, f"Request body is not the expected content: {req.body!r}")
    
        responses.post(
            "http://example.nil/endpoint",
            status=200,
            json={"success": True},
            match=[match_body],
        )
        p = tmp_path / "foo.txt"
        p.write_bytes(CONTENT)
        with requests.Session() as s:
            with p.open("rb") as fp:
>               assert s.post("http://example.nil/endpoint", data=fp) == {"success": True}

test.py:27: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../.local/virtualenvwrapper/venvs/tmp-ca431dc82c2b8f1/lib/python3.12/site-packages/requests/sessions.py:637: in post
    return self.request("POST", url, data=data, json=json, **kwargs)
../../../../.local/virtualenvwrapper/venvs/tmp-ca431dc82c2b8f1/lib/python3.12/site-packages/requests/sessions.py:589: in request
    resp = self.send(prep, **send_kwargs)
../../../../.local/virtualenvwrapper/venvs/tmp-ca431dc82c2b8f1/lib/python3.12/site-packages/requests/sessions.py:703: in send
    r = adapter.send(request, **kwargs)
../../../../.local/virtualenvwrapper/venvs/tmp-ca431dc82c2b8f1/lib/python3.12/site-packages/responses/__init__.py:1173: in send
    return self._on_request(adapter, request, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <responses.RequestsMock object at 0x10b0ffad0>
adapter = <requests.adapters.HTTPAdapter object at 0x10b0dbef0>
request = <PreparedRequest [POST]>, retries = None
kwargs = {'cert': None, 'proxies': OrderedDict(), 'stream': False, 'timeout': None, ...}
request_url = 'http://example.nil/endpoint', match = None
match_failed_reasons = ["Request body is not the expected content: <_io.BufferedReader name='/private/var/folders/l7/wrkq93d133d8zpn36fmqrq0r0000gn/T/pytest-of-jwodder/pytest-91/test_post_file0/foo.txt'>"]
resp_callback = None
error_msg = "Connection refused by Responses - the call doesn't match any registered mock.\n\nRequest: \n- POST http://example.nil...name='/private/var/folders/l7/wrkq93d133d8zpn36fmqrq0r0000gn/T/pytest-of-jwodder/pytest-91/test_post_file0/foo.txt'>\n"

    def _on_request(
        self,
        adapter: "HTTPAdapter",
        request: "PreparedRequest",
        *,
        retries: Optional["_Retry"] = None,
        **kwargs: Any,
    ) -> "models.Response":
        # add attributes params and req_kwargs to 'request' object for further match comparison
        # original request object does not have these attributes
        request.params = self._parse_request_params(request.path_url)  # type: ignore[attr-defined]
        request.req_kwargs = kwargs  # type: ignore[attr-defined]
        request_url = str(request.url)
    
        match, match_failed_reasons = self._find_match(request)
        resp_callback = self.response_callback
    
        if match is None:
            if any(
                [
                    p.match(request_url)
                    if isinstance(p, Pattern)
                    else request_url.startswith(p)
                    for p in self.passthru_prefixes
                ]
            ):
                logger.info("request.allowed-passthru", extra={"url": request_url})
                return self._real_send(adapter, request, **kwargs)  # type: ignore
    
            error_msg = (
                "Connection refused by Responses - the call doesn't "
                "match any registered mock.\n\n"
                "Request: \n"
                f"- {request.method} {request_url}\n\n"
                "Available matches:\n"
            )
            for i, m in enumerate(self.registered()):
                error_msg += "- {} {} {}\n".format(
                    m.method, m.url, match_failed_reasons[i]
                )
    
            if self.passthru_prefixes:
                error_msg += "Passthru prefixes:\n"
                for p in self.passthru_prefixes:
                    error_msg += f"- {p}\n"
    
            response = ConnectionError(error_msg)
            response.request = request
    
>           raise response
E           requests.exceptions.ConnectionError: Connection refused by Responses - the call doesn't match any registered mock.
E           
E           Request: 
E           - POST http://example.nil/endpoint
E           
E           Available matches:
E           - POST http://example.nil/endpoint Request body is not the expected content: <_io.BufferedReader name='/private/var/folders/l7/wrkq93d133d8zpn36fmqrq0r0000gn/T/pytest-of-jwodder/pytest-91/test_post_file0/foo.txt'>

../../../../.local/virtualenvwrapper/venvs/tmp-ca431dc82c2b8f1/lib/python3.12/site-packages/responses/__init__.py:1100: ConnectionError
=========================== short test summary info ============================
FAILED test.py::test_post_file - requests.exceptions.ConnectionError: Connect...
============================== 1 failed in 0.18s ===============================
```

mrob95:
Fixes #719 

Currently I've only dealt with file-like objects and ignored any of the other types that urllib3 supports (see https://github.com/urllib3/urllib3/blob/abbfbcb1dd274fc54b4f0a7785fd04d59b634195/src/urllib3/util/request.py#L220). This seems like an easy improvement since the alternative is keeping file objects in the call list, which are likely to be closed by the time the user gets to them. I don't know whether `responses` should go further though and try to convert every possible request body into a single format (probably `bytes`) to make it easier for the user to assert on.

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -1064,7 +1064,8 @@ class RequestsMock:
         # See GH #719
         if isinstance(body, str) or isinstance(body, bytes) or body is None:
             return body
-        # Based on https://github.com/urllib3/urllib3/blob/abbfbcb1dd274fc54b4f0a7785fd04d59b634195/src/urllib3/util/request.py#L220
+        # Based on
+        # https://github.com/urllib3/urllib3/blob/abbfbcb1dd274fc54b4f0a7785fd04d59b634195/src/urllib3/util/request.py#L220
         if hasattr(body, "read"):
             return body.read()  # type: ignore[attr-defined]
         return body  # type: ignore[no-any-return]
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -1066,9 +1066,9 @@ class RequestsMock:
             return body
         # Based on
         # https://github.com/urllib3/urllib3/blob/abbfbcb1dd274fc54b4f0a7785fd04d59b634195/src/urllib3/util/request.py#L220
-        if hasattr(body, "read"):
-            return body.read()  # type: ignore[attr-defined]
-        return body  # type: ignore[no-any-return]
+        if hasattr(body, "read") or isinstance(body, BufferedReader):
+            return body.read()
+        return body
 
     def _on_request(
         self,
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -1054,6 +1054,21 @@ class RequestsMock:
             params[key] = values
         return params
 
+    def _read_filelike_body(
+        self, body: Union[str, bytes, BufferedReader, None]
+    ) -> Union[str, bytes, None]:
+        # Requests/urllib support multiple types of body, including file-like objects.
+        # Read from the file if it's a file-like object to avoid storing a closed file
+        # in the call list and allow the user to compare against the data that was in the
+        # request.
+        # See GH #719
+        if isinstance(body, str) or isinstance(body, bytes) or body is None:
+            return body
+        # Based on https://github.com/urllib3/urllib3/blob/abbfbcb1dd274fc54b4f0a7785fd04d59b634195/src/urllib3/util/request.py#L220
+        if hasattr(body, "read"):
+            return body.read()  # type: ignore[attr-defined]
+        return body  # type: ignore[no-any-return]
+
     def _on_request(
         self,
         adapter: "HTTPAdapter",
@@ -1067,6 +1082,7 @@ class RequestsMock:
         request.params = self._parse_request_params(request.path_url)  # type: ignore[attr-defined]
         request.req_kwargs = kwargs  # type: ignore[attr-defined]
         request_url = str(request.url)
+        request.body = self._read_filelike_body(request.body)
 
         match, match_failed_reasons = self._find_match(request)
         resp_callback = self.response_callback
--- a/responses/tests/test_responses.py
+++ b/responses/tests/test_responses.py
@@ -1,6 +1,7 @@
 import inspect
 import os
 import re
+import tempfile
 import warnings
 from io import BufferedReader
 from io import BytesIO
--- a/responses/tests/test_responses.py
+++ b/responses/tests/test_responses.py
@@ -2708,3 +2708,23 @@ def test_request_object_attached_to_exception():
 
     run()
     assert_reset()
+
+
+def test_file_like_body_in_request():
+    """Validate that when file-like objects are used in requests the data can be accessed
+    in the call list. This ensures that we are not storing file handles that may be closed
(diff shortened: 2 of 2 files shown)


markstory:
Looking good, but there are a few CI failures due to missing imports.

mrob95:
> Looking good, but there are a few CI failures due to missing imports.

Ah whoops, I think I have fixed the missing import and the line length lint failure now.

PR #739, opened 2024-09-12
Export names `matchers` and `Response` from RequestsMock instances

filbranden:
This makes it possible to import or access only RequestsMock (possibly via a pytest fixture) and still be able to easily access the matchers or assemble a Response object.

It also makes it easier to adopt pytest-responses which uses `responses` as the fixture variable which would conflict with an import of this module.

Add test cases that mask the existing imports and still access matchers or assemble a Response object via the RequestsMock instance.

Also recommend using pytest-responses for a pytest fixture. Rather than explaining how to roll your own in a Markdown snippet, recommend using the companion package to get a standard `responses` pytest fixture easily usable in test cases.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -899,16 +899,19 @@ Integration with unit test frameworks
 Responses as a ``pytest`` fixture
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Use the pytest-responses package to export ``responses`` as a pytest fixture.
+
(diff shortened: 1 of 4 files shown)


filbranden:
See somewhat related getsentry/pytest-responses#30 proposing some usability improvements in pytest-responses as well.

markstory:
Thank you 🎉

PR #742, opened 2024-12-04
feat(release): Replace release bot with GH app

Jeffreyhung:
More details: https://www.notion.so/sentry/DACI-Replace-GitHub-bot-accounts-with-GitHub-Apps-getsentry-release-15109965d1204a91b9be71c49e8b66e0?pvs=4

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,15 +15,22 @@ jobs:
     runs-on: ubuntu-latest
     name: 'Release a new version'
     steps:
+      - name: Get auth token
+        id: token
+        uses: actions/create-github-app-token@5d869da34e18e7287c1daad50e0b8ea0f506ce69 # v1.11.0
+        with:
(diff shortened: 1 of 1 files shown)


beliaev-maksim:
@Jeffreyhung that is not very convenient when notes for open source project are published in internal source. That lowers the visibility on decisions

Jeffreyhung:
@beliaev-maksim appreciate the feedback, we are making this change across 70+ repos and I didn't find a suitable place in our GitHub to put it, so I put it in Notion to save some duplicate contents across the 70+ repos.

But for more context, the `getsentry-release` is a GitHub bot account that is used in various automation, and the `${{ secrets.GH_RELEASE_PAT }}` is a personal access token from that bot account. We are using a regular GitHub account as a bot while it should be a non-human account since there are no humans behind it. Hence, we are replacing it with a GitHub App. Functionality wise, there will be no difference.

beliaev-maksim:
That makes sense, thanks for the context!

PR #743, opened 2024-12-12
Support fuller files kwarg for requests.post

semperos:
Thanks for taking the time to review this! I ran into this type mismatch with some code that leverages the requests library and was hoping to update responses to match.

---

The `files` keyword argument for the `requests.post` function has this type:

```py
files: _Files | None = ...
```

This commit adds this type alias as found in requests version 2.32.2 and updates this library's `multipart_matcher` function's files keyword argument to be of type `_Files`.

### DIFF
--- a/responses/matchers.py
+++ b/responses/matchers.py
@@ -4,7 +4,6 @@ import re
 from json.decoder import JSONDecodeError
 from typing import Any
 from typing import Callable
-from typing import Iterable
 from typing import List
 from typing import Mapping
 from typing import MutableMapping
@@ -16,6 +15,7 @@ from urllib.parse import parse_qsl
 from urllib.parse import urlparse
 
 from requests import PreparedRequest
+from requests.sessions import _Files
 from urllib3.util.url import parse_url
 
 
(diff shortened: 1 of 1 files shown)


beliaev-maksim:
should we just import it from `requests` then ?

semperos:
I neglected to run `tox` before pushing that last commit, there are errors; I'll fix them before pushing again, no need to approve the workflows.

semperos:
I believe Python's `import` prevents importing names with a leading underscore; I get an `ImportError: cannot import name '_Files' from 'requests.sessions'` when I attempt to `from requests.sessions import _Files`.

Two paths forward that I see:

1. Ask the requests maintainers to expose this type.
1. Recreate the types here.

The original PR workflow run showed that the syntax I ported from requests of `tuple[_FileName, _FileContent]` breaks in Python 3.8. If we go with option 2, I can adjust the type syntax to make the code compatible with Python 3.8.

beliaev-maksim:
can you please send a **permalink** to the piece of code at `requests` where they define the type?

semperos:
Apologies for not double-checking the source of the types, I had jumped to them in my editor and didn't realize they weren't provided by the requests implementation itself. The types I included in my first commit were from typeshed.

Here is the permalink to where `_Files` is used: https://github.com/python/typeshed/blob/7cae61f61a042a772010274dd1d8f1ee4454e5a4/stubs/requests/requests/sessions.pyi#L148

Here is the permalink to the definition of `_Files` itself: https://github.com/python/typeshed/blob/7cae61f61a042a772010274dd1d8f1ee4454e5a4/stubs/requests/requests/sessions.pyi#L92

beliaev-maksim:
so, if we add a dependency and import the type, will it work?

PR #745, opened 2024-12-26
Recorder without decorator

niccokunzmann:
I would like to use the recorded in the repl and record my requests.

This adds the code, tests and documentation for this.


markstory:
Thank you 🎉

PR #771, opened 2025-07-21
Fix recording content_type

### LINKED ISSUE #770: Request recorder forgets to record content type
Looking at [the source of the recorder[(https://github.com/getsentry/responses/blob/9ea86fe6af7626e7b270e128b1e8c521c013d948/responses/_recorder.py#L150C39-L156C10), it does not record the content_type of the actual request, and instead synthesizes [it to `"text/plain; charset=utf-8"` instead]("text/plain; charset=utf-8").

This is wrong, as we loose the `content_type` `image/tiff` that the server sends and which we require.

Hour workaround is fairly ugly right now, and looks approximately like this:

```python
        if (len(responses.registered())>2 and
                responses.registred()[1].url.endswith("something_specific")):
            responses.registered()[1].content_type="image/tiff"
```

We would be happy to provide a pull request, please indicate if you would like to take it.

dwt:
The intiial code skipped reorcding the content type and instead fell back to guessing. Now it only falls back to guessing if no explicit content type is provided.

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Other

## Related Issues
- Closes #770

### PR checklist
- [x] Ran `tox` and `pre-commit` checks locally

- [x] Yes
- [ ] No, and this is why: 

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,4 @@
-0.25.8 (in development)
+0.25.8
 ------
 
 * Fix bug where the content type is always recorded as either text/plain or application/json. See #770
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,8 @@
+0.25.8 (in development)
+------
+
+* Fix bug where the content type is always recorded as either text/plain or application/json. See #770
+
 0.25.7
 ------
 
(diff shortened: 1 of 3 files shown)


dwt:
From the commit history it looks like you may have been able to solve that. Any news when we can expect you to cut a new release? I would be quite happy to remove the workarounds we have in our code.

PR #774, opened 2025-07-22
Allow asserts on add_callback() matches

### LINKED ISSUE #727: Cannot match with callback
### Describe the bug

When using standard responses you can use the matching behaviour but the callback API doesn't currently allow you to assert any calls to this callback 

### Additional context

_No response_

### Version of `responses`

0.25.3

### Steps to Reproduce

1. Register callback
2.API currently returns None

### Expected Result

API returns response we can then use with assert

### Actual Result

The api returns none so we can't assert against it

kleptog:
Fixes #727

- [ ] Refactor
- [X] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Other

## Description

It's useful to be able to use the return value of `responses.add()` for asserts. This patch makes this also work for `responses.add_callback()`.

## Related Issues

- Closes #

### PR checklist
- [X] Read the [contributing guidelines](https://github.com/getsentry/responses?tab=readme-ov-file#contributing)
- [X] Ran `tox` and `pre-commit` checks locally
- [X] Added my changes to the [CHANGES](./../CHANGES) file

- [X] Yes
- [ ] No, and this is why: 

### DIFF
--- a/responses/__init__.py
+++ b/responses/__init__.py
@@ -990,20 +990,19 @@ class RequestsMock:
         self.start()
         return self
 
-    def __exit__(self, type: Any, value: Any, traceback: Any) -> bool:
+    def __exit__(self, type: Any, value: Any, traceback: Any) -> None:
         success = type is None
         try:
             self.stop(allow_assert=success)
         finally:
             self.reset()
-        return success
 
     @overload
     def activate(self, func: "_F" = ...) -> "_F":
         """Overload for scenario when 'responses.activate' is used."""
 
     @overload
-    def activate(  # type: ignore[misc]
+    def activate(
         self,
         *,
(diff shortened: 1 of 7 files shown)


responses/__init__.py
```diff
@@ -968,7 +968,7 @@ def add_callback(
         content_type: Optional[str] = "text/plain",
         match: "_MatcherIterable" = (),
     ) -> None:
```
markstory, line 970:
```suggestion
    ) -> BaseResponse:
```

The type annotations will need to be updated.

iamrajjoshi (approved):

markstory:
Thank you 🎉

PR #791, opened 2026-03-30
fix: remove content-type from headers in _add_from_file to avoid RuntimeError

### LINKED ISSUE #741: Inconsistent Behavior of responses._recorder - content_type and Content-Type Header Conflict
### Describe the bug

**Summary:**
When using the _recorder functionality from the responses library to generate YAML files for testing, the recorded output includes both content_type and Content-Type in headers. This results in a RuntimeError during playback due to conflicting definitions, making it difficult to use recorded responses directly in tests without manual intervention.

### Additional context

This is an example recorded response

```yml
# ping_response.yml
responses:
- response:
    body: "{\"message\": \"Test response\"}"
    content_type: text/plain
    headers:
      content-length: '28'
      content-type: application/json; charset=UTF-8
    method: GET
    status: 200
    url: https://dev.opensearch:9200/
```

```bash
poetry show | grep responses
responses                  0.25.3
```

### Version of `responses`

0.25.3

### Steps to Reproduce

**Recording**
```python
@_recorder.record(file_path=OUTPUT_DIR / "get_ping_opensearch_success.yml")
def get_ping_opensearch():
    endpoint = BASE_URL + OPENSEARCH_PORT
    headers = {
        "Content-Type": "application/json",
        "osd-xsrf": "osd-fetch",
        "Accept": "*/*",
        "securitytenant": "global",
    }
    response = requests.request(headers=headers, auth=auth, verify=False, method="get", url=endpoint)
    print(f"Status Code: {response.status_code}")
    pprint(f"Response: {response.headers}")
```

**Output**
```bash
Status Code: 200
("Response: {'content-type': 'application/json; charset=UTF-8', "
 "'content-length': '568'}")
```

**add from file**
```python
@responses.activate
    def test_ping(self):
    responses._add_from_file(file_path="ping_response.yml")
    pass
```
**Error**
```bash
RuntimeError: You cannot define both `content_type` and `headers[Content-Type]`. Using the `content_type` kwarg is recommended.
```

### Expected Result

Not raise `RuntimeError`

### Actual Result

`RuntimeError: You cannot define both `content_type` and `headers[Content-Type]`. Using the `content_type` kwarg is recommended.`

juliosuas:
## Summary

Fixes #741.

The recorder stores `content_type` as a top-level YAML key **and** may also capture a `content-type` entry inside the `headers` dict (since HTTP servers return `Content-Type` as a response header). When `_add_from_file()` calls `add()` with both `content_type=` and `headers={'content-type': ...}`, it raises:

```
RuntimeError: You cannot define both `content_type` and `headers[Content-Type]`.
Using the `content_type` kwarg is recommended.
```

## Fix

Strip any `content-type` key from `headers` before calling `add()` when `content_type` is also present in the loaded YAML. The `content_type` kwarg takes precedence (already recommended by the library).

```python
if headers is not None and "content_type" in rsp:
    headers = {k: v for k, v in headers.items() if k.lower() != "content-type"}
```

This lets recorded YAML files be used in tests without any manual editing.

### DIFF
--- a/responses/tests/test_recorder.py
+++ b/responses/tests/test_recorder.py
@@ -251,6 +251,9 @@ class TestReplay:
         Using mismatched values (``text/html`` in headers vs ``application/json``
         in ``content_type``) ensures the assertion is non-trivial and confirms
         that ``content_type`` takes precedence over the header value.
+
+        The fixture is saved as a ``.yaml`` file so that ``_add_from_file``
+        selects the YAML loader by extension.
         """
         data = {
             "responses": [
@@ -285,12 +288,13 @@ class TestReplay:
             ]
         }
 
-        with open(self.out_file, "w") as f:
+        yaml_file = Path(str(self.out_file) + ".yaml")
+        with open(yaml_file, "w") as f:
             yaml.dump(data, f)
 
         @responses.activate
         def run():
-            responses._add_from_file(file_path=self.out_file)
+            responses._add_from_file(file_path=yaml_file)
 
             # Verify responses were registered without RuntimeError
             assert len(responses.registered()) == 2
--- a/responses/tests/test_recorder.py
+++ b/responses/tests/test_recorder.py
@@ -238,3 +238,63 @@ class TestReplay:
             assert responses.registered()[3].content_type == "text/plain"
 
         run()
+
+    def test_add_from_file_content_type_in_headers(self):
+        """Fixture files may contain Content-Type in both headers and content_type.
+
+        The recorder captures ``Content-Type`` inside the ``headers`` dict *and*
+        as the dedicated ``content_type`` field.  Passing both to ``add()``
+        raises a ``RuntimeError`` because ``content_type`` and a ``Content-Type``
+        header conflict.  ``_add_from_file`` should strip the duplicate header
+        entry so that the dedicated ``content_type`` kwarg wins.
+        """
+        data = {
+            "responses": [
+                {
+                    "response": {
+                        "method": "GET",
+                        "url": "http://example.com/api",
+                        "body": '{"status": "ok"}',
+                        "status": 200,
+                        "headers": {"Content-Type": "application/json"},
+                        "content_type": "application/json",
+                        "auto_calculate_content_length": False,
+                    }
+                },
+                {
+                    "response": {
+                        "method": "POST",
+                        "url": "http://example.com/submit",
+                        "body": "created",
+                        "status": 201,
+                        "headers": {
+                            "Content-Type": "text/plain",
+                            "X-Request-Id": "abc123",
+                        },
+                        "content_type": "text/plain",
+                        "auto_calculate_content_length": False,
+                    }
+                },
+            ]
+        }
+
+        with open(self.out_file, "w") as f:
+            yaml.dump(data, f)
+
+        @responses.activate
+        def run():
+            responses._add_from_file(file_path=self.out_file)
+
+            # Verify responses were registered without RuntimeError
+            assert len(responses.registered()) == 2
+
+            assert responses.registered()[0].url == "http://example.com/api"
+            assert responses.registered()[0].content_type == "application/json"
+
+            assert responses.registered()[1].url == "http://example.com/submit"
+            assert responses.registered()[1].content_type == "text/plain"
+            # Non-content-type headers should be preserved
+            resp = requests.post("http://example.com/submit")
+            assert resp.headers["X-Request-Id"] == "abc123"
+
+        run()
--- a/responses/tests/test_recorder.py
+++ b/responses/tests/test_recorder.py
@@ -190,6 +190,12 @@ class TestReplay:
         if self.out_file.exists():
             self.out_file.unlink()
 
+        # Clean up any extension variants created by individual tests
+        for suffix in (".yaml", ".toml"):
+            p = Path(str(self.out_file) + suffix)
+            if p.exists():
+                p.unlink()
+
         assert not self.out_file.exists()
 
     @pytest.mark.parametrize("parser", (yaml, tomli_w))
--- a/responses/tests/test_recorder.py
+++ b/responses/tests/test_recorder.py
@@ -247,6 +247,10 @@ class TestReplay:
         raises a ``RuntimeError`` because ``content_type`` and a ``Content-Type``
         header conflict.  ``_add_from_file`` should strip the duplicate header
         entry so that the dedicated ``content_type`` kwarg wins.
+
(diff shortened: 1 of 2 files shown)


juliosuas:
One thing to note: this only affects files loaded via `_add_from_file()` — not programmatic `responses.add()` calls which already raise the `RuntimeError` by design (the error message says `content_type` kwarg is recommended). The fix just makes recorded YAML files usable out-of-the-box without manual editing.

markstory (commented):
It would be good to have a test covering this scenario. I could help with the test if you can provide an example payload that is currently not working that will work after this change.

juliosuas:
Added a regression test in `tests/test_recorder.py` — `test_add_from_file_content_type_in_headers`.

It creates a YAML fixture with `Content-Type` in the `headers` dict (which is exactly what the recorder produces) and verifies two cases:
1. When `Content-Type` is the *only* header — after stripping it, headers correctly becomes `None`
2. When there are other headers alongside `Content-Type` — the other headers survive, only `Content-Type` is removed

The test fails with `RuntimeError` before the fix and passes after, confirming it catches the exact regression. Let me know if you'd like additional coverage.

responses/tests/test_recorder.py
```diff
@@ -238,3 +238,63 @@ def _parse_resp_f(file_path):
...
+                    }
+                },
+                {
+                    "response": {
+                        "method": "POST",
+                        "url": "http://example.com/submit",
+                        "body": "created",
+                        "status": 201,
+                        "headers": {
+                            "Content-Type": "text/plain",
```
markstory:
```suggestion
                            "Content-Type": "text/html",
```

Having a different value for `headers` and `content_type` will better capture the coalescing behavior mentioned in the comment.

juliosuas:
Good catch @markstory — updated the test to use `text/html` in `headers['Content-Type']` and `application/json`/`text/plain` in `content_type`. Now the assertion is non-trivial: it confirms `content_type` wins over the conflicting header value rather than passing trivially when both carry the same string.

markstory (approved):

responses/tests/test_recorder.py
```diff
@@ -238,3 +238,71 @@ def _parse_resp_f(file_path):
...
+                },
+            ]
+        }
+
+        with open(self.out_file, "w") as f:
+            yaml.dump(data, f)
+
+        @responses.activate
+        def run():
+            responses._add_from_file(file_path=self.out_file)
```
markstory: You'll either need to save your file as toml, or switch the loader to yaml for this test.

juliosuas:
Good catch @markstory — added `.yaml` extension to the fixture file so `_add_from_file` correctly selects the YAML loader by extension. The fix is minimal: `yaml_file = Path(str(self.out_file) + ".yaml")` and passing that path to `_add_from_file`.

juliosuas:
Fixed both issues flagged by Cursor Bugbot:

1. **Stale fixture file** — `teardown_method` now cleans up `.yaml` and `.toml` extension variants in addition to the base `response_record` path.

2. **Precedence assertion** — the current test already uses mismatched values (`headers["Content-Type"]: "text/html"` vs `content_type: "application/json"` for GET, and `"text/html"` vs `"text/plain"` for POST), so the assertion is non-trivial. The bot's comment referenced an earlier version of the test before `text/html` was added per @markstory's suggestion.

juliosuas:
Both issues flagged by Cursor Bugbot have been addressed (teardown cleanup for yaml/toml variants, and the non-trivial precedence assertion with mismatched values). The PR has @markstory's approval and all CI checks should be passing. Ready to merge when you have a moment — thanks!

juliosuas:
Hi @markstory — just wanted to check in on this one. The fix is minimal (strips duplicate `content-type` from headers before calling `add()`), all three checks pass (Cursor Bugbot ✅, Seer Code Review ✅, Semgrep ✅), and you've already approved it. Is there anything else needed before merge?

markstory:
>  Is there anything else needed before merge?

CI is failing.

juliosuas:
Pushed a fix for the red CI. Root cause turned out to be test state leakage, not the Content-Type logic itself:

`test_add_from_file[tomli_w]` monkey-patches `responses.mock._parse_response_file` to a TOML loader and never restores it. When `test_add_from_file_content_type_in_headers` ran after it, the TOML parser was still installed on the singleton, so the YAML fixture blew up with `tomllib.TOMLDecodeError: Expected '=' after a key in a key/value pair`. The earlier attempts (renaming the fixture to `.yaml`, adding extension cleanup in teardown) didn't help because `_parse_response_file` always goes through `yaml.safe_load` regardless of extension — the dispatch is on the callable, not the filename.

Fix in `teardown_method`: always rebind `responses.mock._parse_response_file` to the class default, so any per-test monkey-patch is contained to that test.

Also picked up a black complaint in `__init__.py` — the Content-Type-stripping dict comp landed at 91 chars, reflowed it to stay under 88.

Happy to split this into two commits if you'd prefer the test-isolation fix and the formatting nit separately.

responses/tests/test_recorder.py
```diff
@@ -190,6 +190,20 @@ def teardown_method(self):
...
+            if p.exists():
+                p.unlink()
+
+        # test_add_from_file[tomli_w] monkey-patches responses.mock._parse_response_file
+        # to a TOML loader and never restores it. Reset it here so later tests
+        # don't inherit the TOML parser when they write YAML fixtures.
+        cls = type(responses.mock)
+        responses.mock._parse_response_file = (  # type: ignore[method-assign]
+            cls._parse_response_file.__get__(responses.mock)
+        )
```
markstory: Wouldn't it be better to unwind the mock in the test that is doing the monkey patching?

juliosuas:
Good call — refactored in `ac93d8f` to capture the original `_parse_response_file` and restore it in a `finally` block within `test_add_from_file`, instead of the class-wide reset in `teardown_method`. Removes the leaked responsibility from the teardown.

juliosuas:
I re-checked the failing checks so this is easier to review:

- The test matrix itself is green across the supported Python / requests / urllib3 combinations.
- The `linting` failure is currently from mypy on an existing `CallList.__getitem__` override signature in `responses/__init__.py`, not from the recorder/content-type diff in this PR.
- The `warden` check currently returns a GitHub API 404 when fetched through Actions logs, so I can't reproduce an actionable failure from this branch.

The PR diff remains scoped to avoiding conflicting `Content-Type` handling when `_add_from_file` passes both headers and `content_type`.

markstory:
Thanks!

PR #801, opened 2026-06-24
fix: query_param_matcher no longer mutates the caller's params dict

gaoflow:
## Problem

`query_param_matcher` converts `int` and `float` values to strings so they can
be compared against URL query-string values (which are always strings). However,
this conversion was done in-place on the dict reference obtained via
`params_dict = params or {}`, which aliases the original dict when a non-empty
mapping is passed. As a result, every numeric value in the caller's dict is
silently overwritten with its string equivalent:

```python
params = {"page": 1, "ratio": 0.5}
matchers.query_param_matcher(params)
# params is now {"page": "1", "ratio": "0.5"} — mutated!
```

This is surprising behaviour that can cause hard-to-debug failures when the
same dict is inspected or reused after the matcher is created.

## Fix

Copy the mapping into a fresh dict before the normalisation loop:

```python
-    params_dict = params or {}
+    params_dict = dict(params) if params else {}
```

A regression test (`test_query_param_matcher_does_not_mutate_input`) is
included.

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,8 @@
   from recorded files regardless of header name case. Previously, responses recorded from
   servers that send lowercase header names (for example over HTTP/2) kept these redundant
   headers in the generated file.
+* Fixed `query_param_matcher` mutating the caller's params dict when numeric
+  values are provided. See #801
 
 0.26.0
 ------
--- a/responses/matchers.py
+++ b/responses/matchers.py
@@ -215,7 +215,7 @@ def query_param_matcher(
 
     """
 
-    params_dict = params or {}
+    params_dict = dict(params) if params else {}
 
     for k, v in params_dict.items():
         if isinstance(v, (int, float)):
--- a/responses/tests/test_matchers.py
+++ b/responses/tests/test_matchers.py
@@ -371,6 +371,24 @@ def test_query_params_numbers():
     assert_reset()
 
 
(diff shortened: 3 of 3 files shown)


markstory (approved):
Could you also add a note to the CHANGES file about this fix?

binggao1230:
Added the CHANGES entry in `03e670a`.

Verification:

```text
uv run --with pytest --with pytest-asyncio --with requests --with urllib3 --with pyyaml --with types-PyYAML python -m pytest responses/tests/test_matchers.py::test_query_param_matcher_does_not_mutate_input -q
# 1 passed

git diff --check
# passed
```

I first tried the system `python -m pytest`, but the local global pytest plugin stack is broken (`pytest_pylint` missing `tomlkit`, then a pytest/pytest-asyncio version mismatch), so I used an isolated `uv run --with ...` environment for the focused test.

binggao1230:
Rebased onto current `master` and resolved the `CHANGES` conflict while keeping both the upstream #798 entry and the #801 note.

Verification after the rebase:

```bash
uv run --with pytest --with pytest-asyncio --with requests --with urllib3 --with pyyaml --with types-PyYAML python -m pytest responses/tests/test_matchers.py::test_query_param_matcher_does_not_mutate_input -q
git diff --check origin/master...HEAD
```

PR #808, opened 2026-07-26
fix: dump recorder output when decorated call raises

### LINKED ISSUE #705: _recorder doesn't record when there's an exception
### Describe the bug

When decorating a function with recorder, nothing gets recorded if an exception is thrown in the function.

### Additional context

_No response_

### Version of `responses`

0.24.1

### Steps to Reproduce

```python
class MyTestCase(TestCase):

    @_recorder.record(file_path="a.yaml")
    def test_a(self):
        requests.get("https://example.com", verify=False)

    @_recorder.record(file_path="b.yaml")
    def test_b(self):
        requests.get("https://example.com", verify=False)
        self.assertEqual(1, 2)
```

### Expected Result

Requests should be recorded in both tests (files a.yaml and b.yaml should be created)

### Actual Result

Requests are only recorded in a (file b.yaml is NOT created)

deepakganesh78:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Other

## Description

The recorder decorator previously dumped captured responses only after the decorated function returned successfully. If the function raised, control left the `with self` block before `dump_to_file()` ran, and the context manager reset the registry, losing the recording.

Move the dump into the wrapper's `finally` path so responses captured before an exception are persisted. If both the decorated function and the dump fail, log the dump failure and preserve the original function exception. A dump failure still propagates normally when the decorated function succeeds.

## Related Issues

Fixes #705

### PR checklist

- [x] Read the contributing guidelines
- [x] Ran the full test suite, mypy, and pre-commit checks locally
- [x] Added the change to `CHANGES`

## Added/updated tests?

- [x] Yes

Regression coverage verifies:

- a request is written when the decorated function subsequently raises;
- the exact original exception is preserved;
- a secondary dump failure is logged without masking that exception;
- dump failures still propagate after successful decorated calls.

## Validation

- `python -m pytest . --asyncio-mode=auto --cov-report term-missing --cov responses -q` — 232 passed
- `python -m pytest responses/tests/test_recorder.py -q` — 10 passed
- `python -m tox -e mypy` — both mypy passes succeeded
- `python -m pre_commit run --all-files` — all hooks passed
- `responses/_recorder.py` — 100% coverage

### DIFF
--- a/responses/_recorder.py
+++ b/responses/_recorder.py
@@ -1,3 +1,4 @@
+import logging
 from functools import wraps
 from typing import TYPE_CHECKING
 
@@ -28,6 +29,8 @@ from responses import RequestsMock
 from responses import Response
 from responses import _real_send
 from responses.registries import OrderedRegistry
+
+logger = logging.getLogger("responses")
 
 
 def _remove_nones(d: "Any") -> "Any":
@@ -115,12 +118,25 @@ class Recorder(RequestsMock):
             @wraps(function)
             def wrapper(*args: "Any", **kwargs: "Any") -> "Any":  # type: ignore[misc]
                 with self:
-                    ret = function(*args, **kwargs)
-                    self.dump_to_file(
-                        file_path=file_path, registered=self.get_registry().registered
-                    )
-
-                    return ret
+                    function_raised = False
+                    try:
+                        return function(*args, **kwargs)
+                    except BaseException:
+                        function_raised = True
+                        raise
+                    finally:
+                        try:
+                            self.dump_to_file(
+                                file_path=file_path,
+                                registered=self.get_registry().registered,
+                            )
+                        except BaseException:
+                            if not function_raised:
+                                raise
+                            logger.exception(
+                                "Failed to dump recorded responses while handling "
+                                "an exception from the decorated function"
+                            )
 
             return wrapper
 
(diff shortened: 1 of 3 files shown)


responses/tests/test_recorder.py
```diff
@@ -133,6 +133,63 @@ def run():
...
+        custom_recorder = _recorder.Recorder()
+
+        def dump_to_file(*args, **kwargs):
+            raise OSError("dump failure")
+
+        custom_recorder.dump_to_file = dump_to_file  # type: ignore[method-assign]
+
+        @custom_recorder.record(file_path=self.out_file)
+        def run():
+            return None
```
markstory, line 188: Shouldn't success include a captured request?

responses/_recorder.py
```diff
@@ -115,12 +118,25 @@ def deco_record(function: "_F") -> "Callable[..., Any]":
...
-                    ret = function(*args, **kwargs)
-                    self.dump_to_file(
-                        file_path=file_path, registered=self.get_registry().registered
-                    )
-
-                    return ret
+                    function_raised = False
+                    try:
+                        return function(*args, **kwargs)
+                    except BaseException:
```
markstory, line 124: Why use BaseException? Should dump files be written to if there are syntax errors?

PR #812, opened 2026-08-13
fix: strip misleading Content-Encoding header from mocked responses

### LINKED ISSUE #724: _recorder.record captures header 'content-encoding: gzip' but does not return gzip content on replay

Sohel2309:
## Summary

Mocked responses can preserve a `Content-Encoding` header even when the response body is not actually encoded.

This causes requests/urllib3 to attempt decoding the body and fail.

## Fix

Strip `Content-Encoding` from mocked response headers before serving the response.

This covers both direct `responses.add()` usage and responses loaded from fixtures.

## Tests

- Added regression tests for both paths.
- Verified the tests fail before the fix and pass after it.
- Full test suite: 230 passed, 1 pre-existing unrelated failure.
- black, isort, flake8 and mypy pass.

Fixes #724


responses/tests/test_responses.py
```diff
@@ -1441,6 +1441,82 @@ def run():
...
+
+    For more details see https://github.com/getsentry/responses/issues/724
+    """
+
+    @responses.activate
+    def run():
+        responses.add(
+            responses.GET,
+            "https://example.org/",
+            body='{"first_name": true}',
```
markstory: Couldn't the response mock be gzip encoded though?

responses/tests/test_responses.py
```diff
@@ -1441,6 +1441,82 @@ def run():
...
+    fixture = tmp_path / "recorded.yaml"
+    fixture.write_text(
+        """\
+responses:
+- response:
+    auto_calculate_content_length: false
+    body: '{"first_name": true}'
+    content_type: application/json
+    headers:
+      content-encoding: gzip
```
markstory: Why not correct the fixture file?

responses/tests/test_responses.py
Sohel2309: Yes, you're right. I initially handled this too broadly. Genuine gzip-encoded mocks are valid and already work correctly, so stripping Content-Encoding when serving the response would break that use case. I've reverted that part and added a regression test to ensure real gzip-encoded mocks continue to be decoded correctly.

responses/tests/test_responses.py
Sohel2309: You're right. The fixture itself is inconsistent. I also found that the recorder was creating the same inconsistency in-memory by storing requests' already-decompressed body together with the original Content-Encoding header. The fix now removes the header at recording time, while leaving response serving and valid gzip mocks unchanged.

beliaev-maksim: sounds like talking with LLM :sweat_smile:

Sohel2309: Fair point . The explanation in the test was unnecessarily verbose. I'll keep the test focused on the behavior being verified and trim the extra commentary.

Issue #738, opened 2024-09-09
PreparedRequest instances are modified by responses


Colin-b:
### Describe the bug

As you can see in https://github.com/getsentry/responses/blob/b68e51388e08399a8828633910eac49d6fd35e2b/responses/__init__.py#L1067 responses, is modifying PreparedRequest instances to incorporate additional attributes.
This is fine in the context of responses, this is not however when this data is returned to the clients.

Could you modify responses so that the behavior from a client perspective is the one of requests (ie: no additional attributes in objects) ?

Thanks again

### Additional context

In the context of writing tests against HTTP behavior that cannot be reproduced with a real live scenario (such as an HTTP failure response from the server), I am trusting my test suite to be able to handle those scenarios and if I see that a PreparedRequest contains an attribute, I assume it will exists in production code as well the day such issue will occur (could be years after deployment time in such scenarios).

### Version of `responses`

0.25.3

### Steps to Reproduce

```python
import requests
import responses

def test_params_in_prepared_request():
    responses.get("https://the_test.url", status=403)
    response = requests.get("https://the_test.url?param=value")
    try:
        response.raise_for_status()
    except requests.RequestException as e:
        print(e.request.params)  # Without responses, an AttributeError will be raise ('PreparedRequest' object has no attribute 'params')
```

### Expected Result

I expect the same failure as in production in my test run

### Actual Result

The AttributeError is not raised when accessing params on a PreparedRequest

markstory:
> Could you modify responses so that the behavior from a client perspective is the one of requests (ie: no additional attributes in objects) ?

So where would you put the additional state that responses needs and do it in a backwards compatible way?

patrickhodel:
The way these params are modified should be considered a bug:

https://github.com/getsentry/responses/blob/93d321222c79b9b931d770fbd94020bad0294297/responses/__init__.py#L1057-L1058

This means it is not possible to mock an api that returns a JSON list with one item.

beliaev-maksim:
@patrickhodel you started with an issue that you access an attribute that does not exist. In general, any linting tool should have caught it on your side, since `requests` do not have this attribute

your second comment describes something else, completely unrelated. 
**Please provide minimal viable example that shows that `responses` modifies some parameters in the incompatible way.** 

This particular [piece of code](https://github.com/getsentry/responses/commit/b71fc15f96a327d54c7ae72c264561049c6cf44e) exists for more than 4 years and never caused any issue

patrickhodel:
[...]

PR #813, opened 2026-08-16
Fix: responses/__init__.py's _on_request() (~line 1099) mutated the SAME...

M001N:
## Summary
In _on_request(), params/req_kwargs are still set on `request` before calling _find_match() (needed because matchers.py's query_param_matcher/request_kwargs_matcher read them during matching). Immediately after matching completes, a shallow copy (`call_request = copy.copy(request)`) is taken to carry those attributes for the internal call log, and `del request.params` / `del request.req_kwargs` strips them from the live request object before it can flow onward to the caller (via response.request or a raised exception's .request). All Call(...)/self._calls.add(...) construction sites in _on_request were switched to use `call_request` instead of `request`, so `responses.calls[i].request.params`/`.req_kwargs` keep working exactly as documented. Updated README.rst's query-params example and two existing tests (test_matchers.py::test_request_matches_params and test_responses.py's assert_params helper used by test_request_param*) that had been asserting .params on the live resp.request object, to instead read it from responses.calls[...].request, since that behavior was the actual bug.

## Problem
getsentry/responses issue reference: getsentry/responses#738

## Root Cause
responses/__init__.py's _on_request() (~line 1099) mutated the SAME live PreparedRequest instance that requests threads back to the caller (via response.request / exception.request) by doing request.params = ... and request.req_kwargs = kwargs directly on it. Since that mutation happened before matching and was never undone, these responses-internal attributes stayed attached to the request object visible to caller code (e.g. in a raised exception's .request), leaking mock-only internals into what should look like a plain, real PreparedRequest.

## Testing
PASS - full suite: 231 passed, 0 failed (18 pre-existing unrelated deprecation warnings).

## Related Issue
getsentry/responses#738

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -368,7 +368,10 @@ deprecated argument.
         constructed_url = r"http://example.com/test?I+am=a+big+test&hello=world"
         assert resp.url == constructed_url
         assert resp.request.url == constructed_url
-        assert resp.request.params == params
+        # ``params`` is only available on ``responses.calls[i].request``, not on the
+        # actual request/response objects returned to the caller, so that mocked
+        # requests don't expose responses-internal attributes to production code.
+        assert responses.calls[0].request.params == params
 
 By default, matcher will validate that all parameters match strictly.
 To validate that only parameters specified in the matcher are present in original request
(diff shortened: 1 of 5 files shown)


README.rst
```diff
@@ -368,7 +368,10 @@ deprecated argument.
         constructed_url = r"http://example.com/test?I+am=a+big+test&hello=world"
         assert resp.url == constructed_url
         assert resp.request.url == constructed_url
-        assert resp.request.params == params
```
markstory, line 371: This could break userland code. But I don't see how we fix the reported issue without removing this attribute.

PR #816, opened 2026-08-30
fix: match sequence-valued params in query_param_matcher

dylanpulver:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Other

## Description

`query_param_matcher` stringified `int`/`float` values only at the top level of the
mapping, so a sequence-valued param never matched:

```python
params = {"ids": [1, 2]}
responses.get(url, match=[matchers.query_param_matcher(params)])
requests.get(url, params=params)   # ConnectionError
```

`responses` parses the request's query string into strings and groups a repeated key
into a list, collapsing a key seen once to a bare value, so the matcher compared
`{"ids": [1, 2]}` against `{'ids': ['1', '2']}`. Tuples and single-item sequences
failed the same way even with string items. Sequences are now normalized the way that
parser produces them.

Deliberately unchanged: an empty sequence, and `None`, still fail to match a request
that omits the key. `requests` drops both, but dropping the key here would also change
what `strict_match=False` filters on.

Checked with an oracle that reuses one dict for both sides across 78 param shapes and
never reads `matchers.py`: 56 failures on master, 2 after, both the empty-sequence case
above. Reverting only `matchers.py` fails all 5 new cases, so none of them is a pin.
Suite goes 233 → 238 passed; coverage and `mypy` failure sets are identical to master.

## Related Issues

- No open issue. Same code path as #484 / #801.

### PR checklist
- [x] Ran `tox` and `pre-commit` checks locally — `pre-commit run` (all hooks, pinned
      versions) is green and both `mypy` invocations report the same errors as clean
      master; pytest ran on 3.13 only, not the full tox matrix.

## Added/updated tests?

- [x] Yes
- [ ] No, and this is why: 

---
Disclosure: prepared with AI assistance (Claude Code, Claude Opus 5, `claude-opus-5`). I have reviewed the change and the measurements above.


responses/matchers.py
```diff
@@ -222,11 +222,24 @@ def query_param_matcher(
...
+        ``responses`` reads the query string back as strings and groups a
+        repeated key into a list, collapsing a key that appears once to a bare
+        value. Sequences here are normalized the same way so that the dict
+        given to ``requests`` can be reused verbatim as the expectation.
+        """
+        if isinstance(value, (list, tuple)):
+            values = [
+                str(item) if isinstance(item, (int, float)) else item for item in value
+            ]
+            return values[0] if len(values) == 1 else values
```
markstory: Don't you also need to recurse into nested structures? Some one could make multi-level query parameter string data that won't have the same normalization applied?

responses/tests/test_matchers.py
```diff
@@ -405,6 +405,34 @@ def run():
     assert_reset()
 
 
+@pytest.mark.parametrize(
+    "expected_query_params",
+    [
+        {"ids": [1, 2]},
+        {"ids": (1, 2)},
+        {"ids": ["a", 2], "page": 1},
```
markstory, line 413:
```suggestion
        {"ids": ["a", 2], "page": 1},
        {"ids": [1, 2], "attrs": {"counter": 3, "favorites": [1, 2]}},
```

Shouldn't data like this also work?

responses/matchers.py
```diff
@@ -222,11 +222,38 @@ def query_param_matcher(
...
+        if not isinstance(value, (list, tuple)):
+            return _stringify(value)
+
+        values: List[Any] = []
+        for item in value:
+            if item is None:
+                # ``requests`` omits a ``None`` element from the query string.
+                continue
+            if isinstance(item, (list, tuple)):
+                values.extend(_stringify(nested) for nested in item)
```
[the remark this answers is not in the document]
markstory: Seems like a relevant gap.

responses/tests/test_matchers.py
```diff
@@ -405,6 +405,71 @@ def run():
...
+        {"ids": [1, 2], "attrs": {"counter": 3, "favorites": [1, 2]}},
+        {"ids": [1]},
+        {"ids": ("a",)},
```
markstory, line 456: Should we also have a case for
```suggestion
        {"ids": None},
```

Currently this fails with

```
E           Request:
E           - GET https://example.com/
E
E           Available matches:
E           - GET https://example.com/ Parameters do not match. {} doesn't match {'ids': None}
```

markstory:
There is still outstanding feedback on missed edge cases that needs to be addressed.

PR #817, opened 2026-08-31
Fix json_params_matcher raising TypeError on nested vs scalar mismatch (#814)

### LINKED ISSUE #814: Non-strict json_params_matcher raises TypeError on nested object/scalar mismatch

philsong4-ai:
## Summary
Fixes #814. `json_params_matcher` with `strict_match=False` raised `TypeError: argument of type 'int' is not iterable` (instead of reporting a normal mismatch) when the expected params had a scalar at a key but the actual JSON body had a nested object at the same key (and vice versa).

## Root cause
`responses/matchers.py::_filter_dict_recursively` recursed whenever the value in the first dict was a mapping, without checking that the corresponding value in the second dict was also a mapping. The recursive call then did `key in <scalar>`, which raised.

## Fix
Recurse only when **both** values are mappings:
```python
if isinstance(val, dict) and isinstance(dict2[k], dict):
    val = _filter_dict_recursively(val, dict2[k])
```
A type mismatch at a key is now reported as a normal non-match `(False, reason)` instead of crashing.

## Test
- Added `test_json_params_matcher_not_strict_scalar_vs_nested_does_not_raise` in `responses/tests/test_matchers.py`, asserting the matcher returns `(False, ...)` (not raises) in both the scalar-vs-nested and nested-vs-scalar directions.
- Full suite: 230 passed; flake8/black/isort clean.

### DIFF
--- a/responses/matchers.py
+++ b/responses/matchers.py
@@ -31,7 +31,7 @@ def _filter_dict_recursively(
     filtered_dict = {}
     for k, val in dict1.items():
         if k in dict2:
-            if isinstance(val, dict):
+            if isinstance(val, dict) and isinstance(dict2[k], dict):
                 val = _filter_dict_recursively(val, dict2[k])
             filtered_dict[k] = val
 
--- a/responses/tests/test_matchers.py
+++ b/responses/tests/test_matchers.py
@@ -210,6 +210,23 @@ def test_json_params_matcher_not_strict_diff_values():
     assert_reset()
 
 
+def test_json_params_matcher_not_strict_scalar_vs_nested_does_not_raise():
+    # Regression test for #814: a non-strict match where the expected value is a
+    # scalar but the actual value is a nested dict used to raise
+    # `TypeError: argument of type 'int' is not iterable` instead of reporting a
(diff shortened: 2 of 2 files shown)


markstory (commented):
Could you update the `CHANGES` file as well? You'll need to add a new section for `0.26.4`

markstory (approved):

markstory:
Duplicate of #815

PR #818, opened 2026-09-08
Expose the augmented callback request type

adamtheturtle:
Responses adds `params` and `req_kwargs` to callback requests, but types callbacks as `PreparedRequest`, forcing consumers to use `getattr` or suppressions.

This exposes `CallbackRequest` (plus its parameter types) and uses it for callback annotations. It remains a `PreparedRequest` alias at runtime because Responses augments existing requests rather than creating a subclass.

Verified with pre-commit, mypy, and all 233 tests on Python 3.8–3.14.


responses/__init__.py
```diff
@@ -54,6 +56,27 @@
...
+    str,
+    Union[str, int, float, List[Optional[Union[str, int, float]]]],
+]
+
+
+class RequestKwargs(TypedDict, total=False):
+    """Keyword arguments passed to ``requests.adapters.HTTPAdapter.send``."""
+
+    stream: bool
+    timeout: Union[float, Tuple[float, float], Tuple[float, None], None]
```
markstory:
```suggestion
    timeout: Union[float, Tuple[float, Union[float, None]], None]
```

Couldn't we combine the two tuple scenarios with Union?

adamtheturtle: Yes - accepted

responses/__init__.py
```diff
@@ -93,6 +122,10 @@ def __call__(
         ],
         models.Response,
     ]
+else:
+    # Callback requests are ordinary PreparedRequest instances at runtime.
+    # The TYPE_CHECKING definition describes the attributes added below.
+    from requests import PreparedRequest as CallbackRequest
```
markstory, line 128: Why does this need to be special cased? It looks like `CallbackRequest` could be exported all the time alongside `RequestKwargs` and `RequestParams`.

adamtheturtle: `CallbackRequest` is exported at runtime, where it is an alias for `PreparedRequest`.

The conditional is needed because `requests` creates a normal `PreparedRequest`. `responses` then adds `params` and `req_kwargs` to that same object before running matchers and callbacks. A normal `CallbackRequest` subclass would not describe the object’s actual runtime class.

We could introduce a genuinely new runtime type, but `responses` would then need to copy or replace the original request. That risks changing object identity and compatibility with matchers, recorded calls, and `response.request`.

The current approach keeps the real `PreparedRequest` object while giving type checkers an accurate description of the attributes available inside callbacks and matchers.

markstory (approved):

PR #820, opened 2026-09-10
fix: keep interleaved query parameter values

nickita-khylkouski:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Other

## Description

Replace the adjacency-dependent `groupby(parse_qsl(...))` logic with `parse_qs(...).items()`. Repeated values are now retained even when another query key appears between them.

## Related Issues

None.

### PR checklist
- [ ] Ran `tox` and `pre-commit` checks locally

I ran the complete pytest suite (236 passed), all pre-commit hooks, and a package build. Current mypy reports the same three pre-existing errors on an unchanged upstream checkout.

- [x] Yes
- [ ] No, and this is why:


responses/tests/test_responses.py
```diff
@@ -2054,6 +2054,41 @@ def run():
...
+            "key1=&key1=one&key2=three&key1=&key2=four",
+            {"key1": ["", "one", ""], "key2": ["three", "four"]},
+        ),
+        (
+            "key%31=two&key2=three&key1=one",
+            {"key1": ["two", "one"], "key2": "three"},
+        ),
+    ],
+)
+def test_request_param_with_interleaved_keys(params, expected):  # type: ignore[misc]
```
markstory, line 2074: I know there are some existing tests with ignores. Can we remove the `type: ignore` on this one so that I don't have more to clean up later?

PR #819, opened 2026-09-10
fix: avoid duplicate Content-Type for lowercase response headers

nickita-khylkouski:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Other

## Description

Wrap mapping-based response headers in `HTTPHeaderDict` before checking for `Content-Type`. This makes the existing membership checks case-insensitive and avoids adding a second content type when a response or callback supplies a lowercase header.

## Related Issues

- Related to #644

### PR checklist
Before submitting this pull request, I have done the following:
- [x] Read the [contributing guidelines](https://github.com/getsentry/responses?tab=readme-ov-file#contributing)
- [ ] Ran `tox` and `pre-commit` checks locally
- [x] Added my changes to the [CHANGES](./../CHANGES) file

I ran the complete pytest suite (239 passed), all pre-commit hooks, and a package build. Current mypy reports the same three pre-existing errors on an unchanged upstream checkout.

- [x] Yes
- [ ] No, and this is why:
- [ ] I need help with writing tests

PR #822, opened 2026-09-19
Fix callback Content-Type overrides from header mappings

jakezwang:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update
- [ ] Other

## Description

Recognize callback header mappings such as `CaseInsensitiveDict` and `HTTPHeaderDict` when replacing the default `Content-Type`. Previously, these produced `text/plain, application/json`, causing Requests to decode UTF-8 JSON as Latin-1.

Extend the existing callback regression across mapping types and header casing, checking both decoded text and JSON.

## Related Issues

Follow-up to #819 for mapping types other than `dict`.

### PR checklist
- [x] Ran `tox` and `pre-commit` checks locally

Validation: 257 tests passed on Python 3.8–3.14, including an additional urllib3 1.x run on Python 3.10. Both mypy modes, all pre-commit hooks, package build and Twine checks passed.

## Added/updated tests?

- [x] Yes
- [ ] No, and this is why:


responses/tests/test_responses.py
```diff
@@ -708,19 +710,23 @@ def run():
 @pytest.mark.parametrize(
     "header_name", ["Content-Type", "content-type", "CONTENT-TYPE"]
 )
-def test_callback_content_type_dict(header_name):  # type: ignore[misc]
+@pytest.mark.parametrize("headers_type", [dict, CaseInsensitiveDict, HTTPHeaderDict])
```
markstory, line 713: Are these types you typically use, or theoretical scenarios that a clanker has devised?

jakezwang:
These are independently constructed edge cases, not header types from a production incident I encountered. I should have made that provenance clear. The test uses `requests.structures.CaseInsensitiveDict` and urllib3's `HTTPHeaderDict` as concrete Mapping implementations to exercise the callback's Content-Type override and UTF-8 decoding. I do not have a real application example to add.

markstory:
> These are independently constructed edge cases, not header types from a production incident I encountered. 

So the answer to my question is no. Since these edge cases aren't driven by real world usage, I see no reason to increase complexity in this library.

PR #823, opened 2026-09-22
Record requests allowed through by passthrough prefixes

### LINKED ISSUE #755: Using pytest fixture doesn't update calls with request/response on passthru
### Describe the bug

Calls not updated when using version 0.5.1 of pytest-responses. 

### Additional context

The workaround for this is to not use the pytest fixture. Works perfectly when using responses module directly. 

### Version of `responses`

0.25.6

### Steps to Reproduce

```python
import requests
def test_weatherstack(client, responses):
    responses.add_passthru(re.compile("https://api.weatherstack.com/\\w+"))
    
    # Trigger request here
    response = requests.get("https://api.weatherstack.com/historical")
    assert response.status_code == 200 # passes assertion
    responses.calls[0].response # index out of bounds
   

```

### Expected Result

Expect the calls to be populated and updated for passthrus. 

### Actual Result

responses.calls are not updated/captured for passthru

Shubham-Padkonde:
Fixes #755.

Requests allowed through by `add_passthru` (URL prefixes or regular expressions) bypass call recording. Tests using a RequestsMock fixture therefore see an empty or incomplete call history even though the real request completed. Record both returned responses and raised exceptions before returning/re-raising, without changing mock matching precedence or retry behavior.

Four regression cases cover prefix/regex registrations, a real local HTTP server interleaved with a registered mock, and a sender raising ConnectionError. They fail on the original code and pass with this change. The exception instance is preserved and the registered mock's call count remains correct.

Validation on Windows / Python 3.13:
- Full pytest suite with the configured asyncio mode and coverage: 255 passed (18 warnings), 85% total reported coverage.
- After final typing adjustments, all 14 passthrough tests pass.
- Black 23.3.0, isort 5.12.0, flake8 and git diff --check pass for changed Python files.
- Mypy reports three errors in unchanged matchers.py/test_recorder.py; verified the identical three errors on the original HEAD using the same environment. No added type-check errors.
- Other Python versions were not run locally.

Prepared and tested with OpenAI Codex assistance. No independent human review is claimed.


markstory:
> No independent human review is claimed.

Sorry, I don't have capacity to review changes if you haven't either.

